PR #21, opened 2026-05-02
Add irc-server tool tests (closes #13)

### LINKED ISSUE #13: Tests: tools list + join/who/leave
Depends on #12.

- `tools/list` returns the 6 expected tools with the right schemas.
- `channel_join` resolves on JOIN ack; `channel_who` includes the MCP nick afterward.
- Peer joins → `channel_who` reflects them. Peer leaves → gone.
- `channel_leave` parts cleanly.

TeakBuilds:
4 integration tests covering the cases from #13:

- `tools/list` returns 6 tools with correct schemas
- `channel_join` resolves on ack; `channel_who` includes own nick
- peer join/leave reflected in `channel_who`
- `channel_leave` parts cleanly (confirmed via peer observing PART)

Also fixed a pre-existing race in `startMcp` — it was returning before the IRC client registered, causing tools to immediately return `isError: true`. Added a readiness poll so callers get a fully-connected context. This unblocked an intermittent failure in `helpers.test.ts` too.

`test/helpers/peer.ts` gets `leaveChannel` and `waitForPart`.

### DIFF
--- a/test/irc-server.test.ts
+++ b/test/irc-server.test.ts
@@ -1,0 +1,75 @@
+import { describe, it, expect, beforeAll } from 'bun:test'
+import { startErgo, type ErgoContext } from './helpers/ergo.js'
+import { startMcp } from './helpers/mcp.js'
+import { connectPeer } from './helpers/peer.js'
+
+let ergo: ErgoContext | null
+
+describe('irc-server MCP tools', () => {
+  beforeAll(async () => {
+    ergo = await startErgo()
+  })
+
+  it('tools/list returns 6 tools with correct schemas', async () => {
+    if (!ergo) { console.log('skipped — ergo not found'); return }
+    const mcp = await startMcp(ergo, 'list-test-mcp')
+    const { tools } = await mcp.client.listTools()
+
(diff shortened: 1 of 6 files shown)


test/irc-server.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+let ergo: ErgoContext | null
+
+describe('irc-server MCP tools', () => {
+  beforeAll(async () => {
+    ergo = await startErgo()
+  })
+
+  it('tools/list returns 6 tools with correct schemas', async () => {
+    if (!ergo) { console.log('skipped — ergo not found'); return }
```
AlexSc: Do we have a quicker way to skip a suite or mark it pending if ergo is unavailable?

AlexSc (changes requested):

test/irc-server.test.ts
TeakBuilds: Done — exported isErgoAvailable() (synchronous binary check) from the ergo helper and switched both test files to describe.if(isErgoAvailable()). Per-test guards are gone.

AlexSc (approved):

PR #34, opened 2026-05-03
feat: add channel_list tool

### LINKED ISSUE #6: Add command to list joined channels
Sonnet in particular seems to lose the thread

TeakBuilds:
closes #6

adds `channel_list` — returns sorted list of currently joined channels from the existing `channelUsers` map. no-arg tool, consistent with the `channel_*` naming convention.

also bumps the tools/list count assertion 6→7 and adds a test that verifies join/leave round-trips.

## Test plan
- [x] `bun test` passes (26/26)

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -429,6 +429,15 @@ mcp.setRequestHandler(ListToolsRequestSchema, async () => ({
         required: ['channel'],
       },
     },
+    {
+      name: 'channel_list',
+      description: 'List all channels currently joined by this MCP instance.',
+      inputSchema: {
+        type: 'object',
+        properties: {},
+        required: [],
+      },
+    },
   ],
 }))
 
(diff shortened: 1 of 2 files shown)


AlexSc (commented):
Is there a way to ask the IRC server directly for what channels we're in? Working off of a local cache feels a little brittle, but it could be acceptable. I'm just not really certain how IRC/IRC clients should work.

AlexSc (approved):
One touchup I'd like to see before merging, but I don't have to review. Let's document somewhere that channel_who and channel_list are handled by a local cache and not a network round trip.

Followup issue for post-release, I think it'd be interesting to have telemetry on "how often is our cache getting out of sync". Something like once every ten minutes just ask the server what's up and compare to local cache. I'm not really sure how to run that at operational scale though; we can notify one agent that something went wrong, but I'm not sure how we get back to the project that it's a "needs fix" if it routinely happens in the wild.

PR #45, opened 2026-05-03
fix: channel_join on already-joined channel returns success

### LINKED ISSUE #7: Fix error on joining already joined channel
Claude seems to think it's a timeout

TeakBuilds:
closes #7

When `channel_join` is called on a channel you're already in, the IRC server doesn't send a JOIN ack — so we'd wait for a resolver that never fires and time out after 5s.

Fix: check `channelUsers` before sending JOIN. If we're already tracked as a member, return success immediately. Channel name normalized to lowercase for the lookup (IRC is case-insensitive; ergo sends lowercase in events).

New test: `channel_join on already-joined channel returns success immediately` — joins a channel, joins again, expects no error and completion in under 1s.

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -542,6 +542,9 @@ mcp.setRequestHandler(CallToolRequestSchema, async req => {
     }
     case 'channel_join': {
       const channel = String(args.channel ?? '')
+      if (channelUsers.has(channel.toLowerCase())) {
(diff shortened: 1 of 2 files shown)


AlexSc:
Do we lowercase the autojoined channels as well? I'm not sure if those go through the MCP join command the same way.

AlexSc (approved):

PR #46, opened 2026-05-03
feat: add -P/--permission-mode flag to roost spawn

### LINKED ISSUE #24: roost spawn: allow overriding --permission-mode (default still auto)
\`bin/roost\` hardcodes \`--permission-mode auto\` for spawned sessions (bin/roost:257). Auto mode is Opus 4.7-only, so non-Opus workers degrade to manual permissioning even though the operator may want \`acceptEdits\` (auto-approve writes/edits, still gate Bash and MCP).

Want: \`roost spawn ... -P acceptEdits\` (or similar) to set the permission-mode flag passed to claude. Default stays \`auto\` for back-compat.

Also: claude's last-occurrence override means \`roost spawn ... -- --permission-mode acceptEdits\` happens to work today (the trailing flag wins), but that's load-bearing on argv parsing order — better to expose it as a first-class roost flag.

TeakBuilds:
adds `-P`/`--permission-mode MODE` to `roost spawn`. default stays `auto` — back-compat preserved.

- validates against `auto`, `acceptEdits`, `plan`, `default`, `bypassPermissions`; exits 1 with clear message on unknown value
- updates the status echo + help text
- manually verified: `roost spawn test-24-throwaway --permission-mode acceptEdits` → footer shows `⏵⏵ accept edits on`

closes #24

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -48,7 +48,7 @@ spawn options:
                              prompts. Daemon is killed by 'roost shutdown'.
   --perm-target NICK         Operator nick to DM for permission prompts.
                              Required when --perm-irc is set.
-  -P, --permission-mode MODE Override --permission-mode passed to claude.
+  --permission-mode MODE     Override --permission-mode passed to claude.
                              Default: auto. Known values: auto, acceptEdits,
                              plan, default, bypassPermissions.
   --cwd PATH                 Working directory for the spawned session.
@@ -67,7 +67,7 @@ spawn options:
 
 Sessions launch under \`zsh -l\` (login shell) so .zprofile loads — required
 for RVM, nvm, and other shell-rc-managed toolchains. Default permission mode
-is auto (Opus 4.7 auto-mode classifier); override with -P.
+is auto (Opus 4.7 auto-mode classifier); override with --permission-mode.
 
 Examples:
   roost spawn worker-1 -c '#my-channel' --cwd ~/Dev/myproject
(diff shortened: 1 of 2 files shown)


bin/roost
```diff
@@ -200,6 +205,12 @@ cmd_spawn() {
     echo "error: --perm-irc requires --perm-target NICK" >&2
     exit 1
   fi
+  case "${permission_mode}" in
```
AlexSc: I would prefer not to add our own validation on top of claude. We should validate what claude does with invalid modes and if it gracefully exits cut this.

What I _would_ like to do is default to acceptEdits if model is sonnet or haiku, since only opus supports auto mode. Or make permission mode required for sonnet or haiku. I'm not really tied to any approach, curious what lead-pm thinks.

AlexSc (changes requested):

AlexSc:
lead-pm, did you have thoughts about my earlier comment?

AlexSc:
I meant the second line.

> What I would like to do is default to acceptEdits if model is sonnet or haiku, since only opus supports auto mode. Or make permission mode required for sonnet or haiku. I'm not really tied to any approach, curious what lead-pm thinks.

AlexSc:
Let's roll it in, and re-review once it lands

AlexSc:
Do we ship a skill file for this? Should update any relevant docs if so

AlexSc (approved):

PR #48, opened 2026-05-03
feat(dispatcher): append comment permalink to relay

### LINKED ISSUE #47: dispatcher: append comment/review permalink to relay so agents can read full body
## Problem

When the dispatcher relays a PR comment over IRC, it truncates after ~140 chars (single-line IRC). Agents (lead-pm, workers) see only the first sentence and miss subsequent paragraphs. Live-witnessed on PR #46 — alex left a 2-paragraph comment, lead-pm responded only to paragraph 1, alex had to say "I meant the second line" before lead-pm fetched via \`gh api\`.

## Proposal

Append the comment/review permalink to the relay so agents know to fetch the full body. Mirrors the existing pattern for review-state events:

\`\`\`
PR foo/bar#46 review by AlexSc (CHANGES_REQUESTED) — https://github.com/foo/bar/pull/46#pullrequestreview-XXX
\`\`\`

For comment events, today's relay:

\`\`\`
PR foo/bar#46 comment by AlexSc: I would prefer not to ad
\`\`\`

Should become:

\`\`\`
PR foo/bar#46 comment by AlexSc: I would prefer not to ad… — https://github.com/foo/bar/pull/46#issuecomment-XXX
\`\`\`

Permalink (anchor-fragment) over bare PR URL — multi-comment PRs make the bare URL useless.

## Scope

- Issue-comment relay
- Review-comment relay (the inline-on-line-N variety, also gets truncated today)
- Review summary relay already has this pattern — leave alone

## Acceptance

- Dispatcher relay for PR/issue comments includes the comment permalink
- Truncation indicator (… or similar) so the reader knows there's more
- Tests cover comment + review-comment paths

## Dogfood note

This is a quality-of-life fix that strongly compounds — every wave we've shipped has had at least one back-and-forth where an agent missed truncated context. Tidy fix to upgrade for the beta example.

TeakBuilds:
Agents were missing truncated PR/issue comment bodies — the relay showed only the first line with no way to retrieve the rest.

Appends `comment_url` (already carried in the event) after the snippet, plus `…` when the body is longer than what's shown. Mirrors the existing review-state pattern.

Covers `pr_review_comment`, `pr_conversation_comment`, and `issue_comment`. `pr_review_submitted` already had this pattern and is left alone.

6 new cases in `_run_self_test`; wired into `bun test` via `test/dispatcher.test.ts`.

closes #47

### DIFF
--- a/bin/orchestrator_poll
+++ b/bin/orchestrator_poll
@@ -572,14 +572,20 @@ def format_event(event: dict) -> str:
         path = event.get("path")
         line = event.get("line")
         where = f" at {path}:{line}" if path else ""
-        snippet = (event.get("body_preview") or "").strip().splitlines()[:1]
-        snippet = snippet[0][:160] if snippet else ""
-        return f"PR {tag} comment by {author}{where}: {snippet}"
+        body_preview = (event.get("body_preview") or "").strip()
+        snippet_lines = body_preview.splitlines()[:1]
+        snippet = snippet_lines[0][:160] if snippet_lines else ""
+        ellipsis = "…" if body_preview != snippet else ""
+        url = event.get("comment_url") or event.get("url", "")
+        return f"PR {tag} comment by {author}{where}: {snippet}{ellipsis} — {url}"
     if kind == "issue_comment":
         author = event.get("author") or "?"
-        snippet = (event.get("body_preview") or "").strip().splitlines()[:1]
-        snippet = snippet[0][:160] if snippet else ""
-        return f"Issue {tag} comment by {author}: {snippet}"
+        body_preview = (event.get("body_preview") or "").strip()
+        snippet_lines = body_preview.splitlines()[:1]
+        snippet = snippet_lines[0][:160] if snippet_lines else ""
+        ellipsis = "…" if body_preview != snippet else ""
+        url = event.get("comment_url") or event.get("url", "")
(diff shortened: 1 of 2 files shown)


bin/orchestrator_poll
```diff
@@ -553,6 +553,15 @@ def should_push(event: dict) -> bool:
 # Channel formatting
 # ---------------------------------------------------------------------------
 
+def _comment_snippet(event: dict) -> tuple[str, str, str]:
+    body_preview = (event.get("body_preview") or "").strip()
+    snippet_lines = body_preview.splitlines()[:1]
+    snippet = snippet_lines[0][:160] if snippet_lines else ""
+    ellipsis = "…" if len(body_preview) > len(snippet) or "\n" in body_preview else ""
```
AlexSc: Why not check snippet_lines length instead of searching for "\n"?

AlexSc (changes requested):
Minor code cleanliness, probably worth checking all of _comment_snippet to see if there are other quick wins.

Good job wiring the orchestrator check into the bun test.

bin/orchestrator_poll
TeakBuilds: [worker-47] fixed — removed the `[:1]` slice, now using `len(snippet_lines) > 1`

AlexSc (approved):

PR #49, opened 2026-05-03
fix(spawn): gate PermissionRequest hook on --perm-irc

### LINKED ISSUE #3: Don't blanket load the PermissionRequest hook
This should only be loaded in sessions where we're wiring up IRC permissions

TeakBuilds:
`hooks/hooks.json` was registering the PermissionRequest hook for every session. Now it's only wired for sessions spawned with `--perm-irc`.

- `hooks/hooks.json`: empty `"hooks": {}`
- `bin/roost cmd_spawn`: when `--perm-irc`, writes `/tmp/roost-perm-settings-<nick>.json` with the hook config and passes `--settings <path>` to claude (`--settings` merges with existing settings layers, it doesn't replace)
- `bin/roost stop_permbot`: cleans up the settings file alongside the permbot pid/socket (mirrors existing leak-on-direct-kill behaviour)
- README + SKILL.md: drop "wires PreToolUse" / "safe to have wired at all times" language

closes #3

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,7 @@ When a Claude Code session loads `roost-irc` as an MCP and connects:
 ```
 
 This puts `roost`, `roost-permbot`, and `irc-permission-prompt` on
-your PATH, auto-loads the `roost-irc` MCP for every session, and wires
-the `PreToolUse` hook for IRC permission oversight.
+your PATH and auto-loads the `roost-irc` MCP for every session.
 
 After installing, pull dependencies:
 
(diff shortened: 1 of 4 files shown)


AlexSc (changes requested):
1. Can --settings be called multiple times? If we're using it, then a spawn using --settings horks this.
2. We're creating a lot of temp files at this point. Should we be using mktmp or similar to make sure things really stay isolated? If someone is working on multiple repos with overlapping issue numbers we're in trouble.

AlexSc (changes requested):
Okay, let's actually roll back this mktemp, sorry. lead-pm, can we get a followup issue to clean up how we handle tmp files to ensure we don't get conflicts?

AlexSc (approved):

PR #51, opened 2026-05-03
feat: add roost send command

### LINKED ISSUE #10: Add roost send command
Sometimes it's useful to dm an agent through the harness instead of a channel.

We tell the things that these are different but are they really?

Don't say that part in the docs.

TeakBuilds:
closes #10

`roost send <nick> <message...>` injects text directly into a nick's Claude Code TUI via `tmux send-keys -l` + Enter — same as typing it at the prompt.

- validates session exists (`tmux has-session`), exits 4 if not found
- rejects multi-line messages (each newline would submit mid-input)
- no daemon, no IRC connection

**Manual repro:** `bin/roost send lead-pm "hello"` — message appeared in lead-pm's TUI prompt (exit 0). `bin/roost send nonexistent "hi"` → `no tmux session named 'roost-nonexistent'` (exit 4).

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -387,17 +387,16 @@ cmd_send() {
   local nick="$1"
   shift
   local message="$*"
-  if [[ "${message}" == *$'\n'* ]]; then
-    echo "error: message must be a single line (newlines not allowed)" >&2
-    exit 1
-  fi
   local session_name="${SESSION_PREFIX}${nick}"
   if ! tmux has-session -t "${session_name}" 2>/dev/null; then
     echo "no tmux session named '${session_name}'" >&2
     exit 4
   fi
-  tmux send-keys -t "${session_name}" -l "${message}"
+  local buf="roost-send-${nick}-$$"
+  printf '%s' "${message}" | tmux load-buffer -b "${buf}" -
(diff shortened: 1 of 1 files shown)


AlexSc (changes requested):
I'm sorry, that issue wasn't very clear :(

The idea with `roost send` was to wrap tmux send-keys. Allow for "DM" directly to harness input instead of IRC. It comes up a fair bit.

bin/roost
```diff
@@ -377,6 +378,28 @@ cmd_status() {
...
+    echo "error: send requires a nick and message" >&2
+    echo "usage: roost send <nick> <message...>" >&2
+    exit 1
+  fi
+  local nick="$1"
+  shift
+  local message="$*"
+  if [[ "${message}" == *$'\n'* ]]; then
+    echo "error: message must be a single line (newlines not allowed)" >&2
+    exit 1
```
AlexSc: In a previous iteration on this CLI we passed in the initial prompt file using send-keys and supported newlines. Can you dig through history to see how that worked, and see if we can support newlines here?

AlexSc (changes requested):

AlexSc (approved):

PR #54, opened 2026-05-03
feat(tmp-hygiene): per-session mktemp dir + tmux env discovery

### LINKED ISSUE #52: tmp file hygiene: unify the /tmp/roost-* naming + cleanup story
## Problem

Roost writes several files into /tmp keyed by nick:
- /tmp/roost-permbot-<nick>.sock
- /tmp/roost-permbot-<nick>.pid
- /tmp/roost-permbot-<nick>.log
- /tmp/roost-perm-settings-<nick>.json (introduced by #3 / PR #49)

All naming-by-nick → predictable but collides if (a) two operators on the same machine spawn the same nick, (b) a stale file from a prior session leaks past a non-clean shutdown.

PR #49 briefly tried mktemp + a pointer file for isolation; rolled back per review since it added complexity. Want to revisit holistically rather than per-file.

## Open questions

- Per-user tmpdir (~/.cache/roost/<nick>/...) vs shared /tmp?
- Atomic cleanup on every spawn-of-same-nick (clobber stale before write)?
- Lockfile to detect concurrent same-nick spawns and error early?
- mktemp + pointer redux — was the pointer pattern just unnecessary or actually useful?

## Acceptance

- One coherent scheme covering socket / pid / log / settings (and any future per-nick files)
- Spawn-after-crash-cleanup is a no-op without manual rm
- Same-nick concurrent spawn surfaces a clear error rather than racing
- Doc the convention so future per-session files follow the same pattern

## Background

Surfaced from PR #49 review where alex asked: "If we're using --settings, then a spawn using --settings horks this" — the immediate fix landed as a fixed /tmp path, but the broader hygiene question is worth its own issue.

TeakBuilds:
closes #52

## what

each `roost spawn` creates a `mktemp -d /tmp/roost.XXXXXXXX` dir. all per-session files live there. `ROOST_DATA_DIR` is stored in the tmux session env for shutdown discovery.

kills the cross-repo nick collision problem — two repos both running `issue-18` get separate dirs.

## shape

- `perm_paths(data_dir)` — dir is the input, not nick
- `stop_permbot(data_dir)` — kills daemon + `rm -rf` the whole dir
- `cmd_spawn` — mktemp + trap-EXIT for failure cleanup, `tmux set-environment ROOST_DATA_DIR` after session creation, trap cleared on success
- `cmd_shutdown` — reads `ROOST_DATA_DIR` from tmux session env
- `irc-permission-prompt` — drops deterministic sock fallback; ROOST_PERM_SOCK is always explicit when --perm-irc is used
- `roost-permbot` — DEBUG_LOG defaults to `<sock-dir>/permbot.log`

## discovery

```
tmux show-environment -t roost-<nick> ROOST_DATA_DIR
```

spawn output also prints the dir directly.

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -92,30 +92,34 @@ require_ircd() {
   fi
 }
 
-# Side-daemon (permbot) lifecycle. PERM_SOCK / PERM_PID / PERM_LOG paths
-# are derived from the worker nick — one daemon per worker, lifecycle
-# tied to the tmux session via roost spawn / roost shutdown.
+# Side-daemon (permbot) lifecycle. All per-session files live under a
+# mktemp-generated directory created at spawn time (ROOST_DATA_DIR).
+# The dir is stored in the tmux session env so shutdown can find it:
+#   tmux show-environment -t roost-<nick> ROOST_DATA_DIR
+# Future per-session files belong in that dir; no additional convention needed.
 perm_paths() {
-  local nick="$1"
-  PERM_SOCK="/tmp/roost-permbot-${nick}.sock"
-  PERM_PID="/tmp/roost-permbot-${nick}.pid"
-  PERM_LOG="/tmp/roost-permbot-${nick}.log"
-  PERM_SETTINGS="/tmp/roost-perm-settings-${nick}.json"
+  local data_dir="$1"
+  PERM_DIR="${data_dir}"
+  PERM_SOCK="${data_dir}/permbot.sock"
+  PERM_PID="${data_dir}/permbot.pid"
+  PERM_LOG="${data_dir}/permbot.log"
+  PERM_SETTINGS="${data_dir}/perm-settings.json"
 }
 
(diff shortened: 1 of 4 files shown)


bin/roost
```diff
@@ -225,6 +229,16 @@ cmd_spawn() {
     done
   fi
 
+  # Create a per-session data directory. mktemp guarantees uniqueness across
+  # concurrent spawns and across repos using identical nick names.
+  local ROOST_DATA_DIR
+  ROOST_DATA_DIR="$(mktemp -d /tmp/roost.XXXXXXXX)"
```
AlexSc: Let's include the nick in our baseline name too.

AlexSc (changes requested):
Also worth a line in the SKILL.md or somewhere to direct the agent on how to find logs for the permbot if needed.

AlexSc (approved):

PR #56, opened 2026-05-03
feat(mcp): dedupe chathistory replay on reconnect

### LINKED ISSUE #44: MCP: dedupe history replay on reconnect (within same context window)
## Problem

When the MCP loses its IRC connection (server restart, network blip, operator stepping away) and reconnects, the channel history gets replayed in full. Live-witnessed in #leads-roost-dev: same batch of ~30 messages delivered twice with new seq numbers but `historical="true"`. Floods the agent's context with duplicates it has already processed.

Distinct from #40 (channel-join replay) — same surface symptom, different trigger.

## Scope

Dedupe history replay on reconnect, but **only within the same context window**. If a disconnect spans a compaction, the post-compact context has lost the prior history — backfilling is valuable there, not noise.

Investigation needed — multiple moving parts:
- Is the replay coming from the MCP server, ergo, or both?
- How does the MCP know whether the consuming agent has compacted?
- What's the right key for "same context window"? (process restart? explicit session id? agent-side ack?)
- Interaction with #40 (channel-join replay) — should converge on one dedupe path.

## Acceptance

- New alpha issue, kept open for investigation
- Proposal posted before code: where dedupe lives + how cross-compact backfill is preserved
- Implementation gated on plan review

TeakBuilds:
Closes #44.

## What

Seen-fingerprint set in the MCP suppresses duplicate channel notifications when ergo replays chathistory on reconnect or rejoin. Fingerprint: `channel|sender|server-ts|text[:64]`. Text prefix handles the millisecond-collision case lead-pm flagged.

Also requests the `server-time` cap so live messages carry the same `@time` stamp ergo uses in chathistory batches — fingerprints match across both paths.

## Cross-compaction

MCP writes its PID to `${ROOST_DATA_DIR:-/tmp}/roost-mcp-{nick}.pid` at startup. New `bin/roost-compact-hook` sends SIGUSR1 on PreCompact; SIGUSR1 handler clears the seen-set so the next backfill after reconnect re-delivers messages compacted out of context. Hook wired in `.claude/settings.json`; `.gitignore` updated to track that file while keeping `settings.local.json` local.

Set cap: `HISTORY_SIZE × joined-channels × 2`, evict-oldest on overflow.

## Tests

Three new cases in `test/chathistory.test.ts`:
- live message received → part+rejoin → no duplicate historical notification
- historical message (first join) → part+rejoin → no duplicate
- SIGUSR1 clears seen-set → part+rejoin → message replays correctly

All 32 tests pass.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


.claude/settings.json
```diff
@@ -0,0 +1,14 @@
+{
```
AlexSc: This goes in hooks/hooks, gated on using roost. Same as permbot. Worth a note in CLAUDE.md that since you're working on a plugin some things are in different places.

src/irc-server.ts
```diff
@@ -96,6 +96,39 @@ const pushHistory = (key: string, msg: IrcMessage) => {
   history.set(key, buf)
 }
 
+// ---- Replay dedupe (issue #44) -----------------------------------------
+
+const seenFingerprints = new Set<string>()
+
+const msgFingerprint = (msg: IrcMessage): string =>
```
AlexSc, line 105: Any reason not to use a per-channel cache?

AlexSc (changes requested):
Also rebase on main since datadir landed, clean up "what if no datadir"

AlexSc (approved):

PR #61, opened 2026-05-03
feat(coverage): wire bun test --coverage with thresholds (#19)

### LINKED ISSUE #19: Wire bun test --coverage with thresholds
Depends on #12 (and ideally lands after the other test issues so we have something to measure).

- Add `bunfig.toml` with `[test] coverage = true`, line threshold 0.85, branch threshold 0.75 (per-file, not global).
- `bun test --coverage` outputs text summary and lcov.
- Document in README how to run.
- Legacy chunker code in `src/irc-server.ts` will show as uncovered — fine for now, separate cleanup.

TeakBuilds:
closes #19

`bunfig.toml` with `coverage = true`, text + lcov reporters, global line 0.85 / branch 0.75 thresholds. Issue called for per-file but bun's tooling doesn't expose file-pattern filtering — shipping global with a comment noting the deviation and a follow-up to revisit if bun adds per-file support.

gate is aspirational today: only `src/constants.ts` is imported by tests (100% covered); `src/irc-server.ts` runs as a subprocess and never appears. gate gets real bite as `src/` grows importable modules.

`coverage/` added to .gitignore; README gets a Testing section.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ var/
 
 # Test artifacts
 tests/logs/
+coverage/
 
 .claude/
 
--- a/README.md
+++ b/README.md
@@ -251,8 +251,8 @@ For coverage (line ≥ 85%, branch ≥ 75% globally):
 bun test --coverage
 ```
 
-Coverage only reflects files imported by tests. `src/irc-server.ts` runs as a
-subprocess via the MCP harness and won't appear — that's expected.
+Coverage includes `src/irc-server.ts` via the in-process tests
+(`test/irc-server-inprocess.test.ts`).
 
 ## Known limitations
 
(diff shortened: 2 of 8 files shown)


AlexSc (changes requested):
Can we figure something out for irc-server? More than aspirational, this is borderline useless without actual coverage analysis.

Correct me if I've misunderstood where things are.

AlexSc:
This is still... not a lot. Can you take a look at https://github.com/oven-sh/bun/issues/17867 I think it may have some relevant ideas.

AlexSc (changes requested):
Can we resolve merge conflicts first? And I think this needs another agent review pass given how big it is.

AlexSc (approved):
I'd like a tracking issue before we close out the Alpha milestone to have an Opus agent do a full simplify pass on everything.

And I think I'd like another tracking issue -- probably doesn't need opus -- to move more of our existing tests over to the in-memory mcp if we can. Blocked on the Opus simplify.

PR #69, opened 2026-05-04
feat(unread): passive channel-unread awareness

### LINKED ISSUE #9: Add passive awareness of "unread"
Unread isn't quite right because it's forced to read everything, but some kind of ongoing indication of "you are in these channels" might help.

Keeping it to a post compaction hook could be good enough -- maybe have haiku summarize recent channel activity too?

The problem I'm trying to solve is "it loses track of what it's looking at", because it isn't looking at anything, it's getting random irc messages shoved into context.

TeakBuilds:
Closes #9.

## What

Agents joined to multiple channels now have passive awareness of which channels have pending activity. After compaction, a `PostCompact` hook triggers an **unread-summary** notification listing every channel with unread messages (sender + message preview per channel, enough to triage without context-switching). Clean state gets an "all caught up" confirmation.

**Ack signals** (any of these clears a channel's unread count):
- Outbound `channel_message` or `direct_message` — sending = you handled it
- `channel_history` fetch — explicit read is the strongest ack signal
- New `channel_ack` tool — explicit clear when you've reviewed a channel another way

**Post-join baseline rule:** historical messages replayed on join start at 0. Unread accrues only from messages arriving after the join settles. Documented in code + this PR.

## Key files

- `src/irc-server.ts` — unread Map, `channel_ack` tool, `emitUnreadSummary()`, SIGUSR2 handler, `channel_list` unread annotations
- `bin/roost-post-compact-hook` — PostCompact hook, sends SIGUSR2
- `bin/roost` — wires PostCompact alongside existing PreCompact

## Test plan

- `bun test` — 25 pass, 0 fail
- 6 new in-process tests covering: inbound increments, send/history/ack clears, emitUnreadSummary with and without pending unread, post-join baseline

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -282,8 +282,7 @@ export function createMcpServer(ircClient: any, config: McpServerConfig): { serv
     addFingerprint(msg)
     if (!extras.historical) {
       const prev = unread.get(msg.channel)
-      const preview = msg.text.length > 80 ? msg.text.slice(0, 77) + '...' : msg.text
-      unread.set(msg.channel, { count: (prev?.count ?? 0) + 1, lastSender: msg.sender, lastPreview: preview })
+      unread.set(msg.channel, { count: (prev?.count ?? 0) + 1, lastSender: msg.sender, lastPreview: msg.text })
     }
     const seq = ++receiveSeq
     const meta: Record<string, string> = {
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -482,7 +482,10 @@ export function createMcpServer(ircClient: any, config: McpServerConfig): { serv
         const { chunks, mode } = sendWithSplit(channel, text)
         unread.delete(channel)
         const unreadSuffix = unread.size > 0
-          ? `\nunread: ${[...unread.entries()].map(([ch, i]) => { const p = i.lastPreview.length > 40 ? i.lastPreview.slice(0, 37) + '...' : i.lastPreview; return `${ch} (${i.count}, last: ${i.lastSender}: "${p}")` }).join(', ')}`
+          ? '\nunread:\n' + [...unread.entries()].map(([ch, i]) => {
+              const raw = i.lastPreview.length > 40 ? i.lastPreview.slice(0, 37) + '...' : i.lastPreview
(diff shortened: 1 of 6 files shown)


AlexSc (changes requested):
Can we include "unread channels" in our reply from sending messages? I think that will create the passive awareness that I'm looking for. But what do both of you think -- worker and lead-pm?

src/irc-server.ts
```diff
@@ -482,24 +482,30 @@ export function createMcpServer(ircClient: any, config: McpServerConfig): { serv
         const text = String(args.text ?? '')
         const { chunks, mode } = sendWithSplit(channel, text)
         unread.delete(channel)
+        const unreadSuffix = unread.size > 0
+          ? `\nunread: ${[...unread.entries()].map(([ch, i]) => { const p = i.lastPreview.length > 40 ? i.lastPreview.slice(0, 37) + '...' : i.lastPreview; return `${ch} (${i.count}, last: ${i.lastSender}: "${p}")` }).join(', ')}`
+          : ''
         const note =
```
AlexSc, line 496: Hey... how are y'all feeling about this response format?

AlexSc:
Ah, my question was more general -- the format period. I haven't seen what you see, just curious if you're finding the general response format helpful. But I do think one-channel per line and your other fixes were nice.

For this PR specifically:
- Let's have a helper function for formatting an individual unread line that we can reuse (maybe it takes previewLength as a parameter?)
- Then let's have a helper function to assemble the full unread list/string/post-send-meta.

And also let's get the merge conflicts resolved.

src/irc-server.ts
```diff
@@ -261,6 +280,11 @@ export function createMcpServer(ircClient: any, config: McpServerConfig): { serv
     extras: { buffered?: boolean; chunkCount?: number; historical?: boolean } = {},
   ) => {
     addFingerprint(msg)
+    if (!extras.historical) {
+      const prev = unread.get(msg.channel)
+      const preview = msg.text.length > 80 ? msg.text.slice(0, 77) + '...' : msg.text
```
AlexSc: Since we format on demand, let's axe this write time formatting.

AlexSc (approved):
Minor cleanup nit. Take care of that and I don't need to re-review

PR #70, opened 2026-05-04
feat(dispatcher): relay full comment bodies via draft/multiline IRC

### LINKED ISSUE #66: orchestrator: relay full comment bodies via multiline IRC instead of truncating
The dispatcher (\`bin/orchestrator_poll\` / \`src/orchestrator_poll.ts\`) currently truncates each PR/issue comment to a single IRC line + permalink. Agents (and humans) routinely respond as if they've read the whole thing because the snippet looks self-contained, miss later sentences, and ship half-answers — see e.g. PR #61's "I'd also like another tracking issue" trailing clause that the lead-pm missed because the dispatcher cut after the first sentence.

Now that ergo + the MCP both support IRCv3 \`draft/multiline\` cleanly (PR #51 etc.), the dispatcher should relay the full body as a multiline batch. Permalink stays at the end as a fallback.

## Acceptance

- PR/issue comment relays send the full body as one multiline IRC message
- Permalink line still appears (so a fallback chathistory client without multiline can find the canonical link)
- A pathological body (huge wall of text exceeding ergo's max-lines / max-bytes) gracefully falls back to truncate-with-permalink rather than erroring
- Manual sanity check on a multi-paragraph PR review

[lead-pm]

TeakBuilds:
Fixes #66. See also #72 for full IRC library migration followup.

## What

Comments and reviews were truncated to a snippet + permalink on a single IRC line. Agents (and humans) read the snippet as self-contained and miss later sentences.

Now the dispatcher negotiates `draft/multiline` and relays the full body as a BATCH, with the permalink as the last batch member so non-multiline clients still see the canonical link.

## How

- `IrcClient.connect()`: `CAP LS 302 / REQ / ACK / END` handshake via regex parser; resets per-connection so `reconnect()` re-negotiates
- `IrcClient.comment_privmsg()`: sends `BATCH` if enabled + fits server limits, else falls back to `privmsg(fallback)` with existing truncated-snippet text
- `_utf8_split(text, limit, *, boundary)`: shared core for `_irc_split_text` and `_multiline_split_line`
- `_event_tag(event)`: extracted one-liner used by `format_event` and `_format_comment_header`
- `CAP_MULTILINE = "draft/multiline"` constant; 4 module-level regexes for CAP parsing
- `dispatch_events_irc()`: comment event kinds route to `comment_privmsg()`; non-comment events unchanged
- 16 new self-test checks

All existing tests still pass (`bun test`: 25 pass, 47 skip).

### DIFF
--- a/bin/orchestrator_poll
+++ b/bin/orchestrator_poll
@@ -47,6 +47,7 @@ import subprocess
 import sys
 import tempfile
 import random
+import re
 import time
 import traceback
 from datetime import datetime, timezone
@@ -686,6 +687,15 @@ def _multiline_split_line(line: str) -> list[str]:
     return _utf8_split(line, MULTILINE_LINE_BYTES, boundary=False)
 
 
+# Matches: `:server CAP <nick> <sub> [*] :<caps>`
+# group 1 = subcommand, group 2 = "*" (non-final LS) or None, group 3 = caps string
+_RE_CAP = re.compile(r'^:\S+ CAP \S+ (\w+)(?: (\*))?(?:\s+:?(.*))?')
+_RE_001 = re.compile(r'^:\S+ 001 ')
+# Matches the draft/multiline cap token; group 1 = "key=val,…" params or None
+_RE_MULTILINE_CAP = re.compile(r'(?:^|\s)draft/multiline(?:=(\S+))?(?:\s|$)')
+_RE_MULTILINE_PARAMS = re.compile(r'max-bytes=(\d+)|max-lines=(\d+)')
+
+
 class IrcClient:
     def __init__(self, server: str, port: int, nick: str, default_channels: Iterable[str] = ()):
         self.server = server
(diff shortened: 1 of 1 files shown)


bin/orchestrator_poll
```diff
@@ -669,25 +695,71 @@ class IrcClient:
...
             for line in self._read_lines(timeout=1.0):
-                parts = line.split(" ", 3)
-                if len(parts) >= 2 and parts[0].startswith(":") and parts[1] == "001":
-                    registered = True
-                    break
                 if line.startswith("PING "):
                     self._send_raw("PONG " + line[5:])
+                    continue
+                parts = line.split(" ", 4)
+                if len(parts) >= 4 and parts[1] == "CAP":
```
AlexSc: There has to be a better way to do this parser. This is... insane.

At this point maybe we should be thinking about uvx and asking if there's a python lib that does IRCv3

AlexSc (changes requested):

AlexSc:
We'll do followup issue, for now could the parser at least be a regex or something a little less... nested?

AlexSc (approved):

PR #76, opened 2026-05-04
maint(stubs): delete stub-server.ts and tools-only-stub.ts

### LINKED ISSUE #68: stub-server.ts / tools-only-stub.ts retention review
[reviewer-63] surfaced from #63 simplify pass.

`src/stub-server.ts` (104 LOC) and `src/tools-only-stub.ts` (63 LOC) were built for Test 1 / Test 3 to prove the load-bearing assumptions in `docs/LEARNINGS.md` (claude/channel capability + tools capability coexisting; multi-MCP per session; cache TTL behavior). Tests passed; the production MCP `src/irc-server.ts` exists.

## Reachability

- **No test imports either file.** Grep across `test/` returns zero.
- **`bun stub` script** in `package.json` invokes `bun run src/stub-server.ts` — but this just starts the stub MCP standalone; nothing runs it programmatically.
- README's `Layout` section names both files. `docs/LEARNINGS.md` references them in the historical test write-up.

## What `bun stub` actually demonstrates

Running it produces a stdio MCP that emits a tick channel notification every 60s and exposes an `echo` tool. Useful as a hand-driven probe to verify a Claude Code session sees `notifications/claude/channel` events at all — i.e., a smoke test for the channel surface independent of IRC. Nothing automated consumes it.

## Recommendation: kill

The load-bearing assumptions are confirmed and the production code is the proof. The stubs aren't fixtures, aren't covered, and the diagnostic role they play (probe `claude/channel` end-to-end) can be reproduced by spawning a real `roost-irc` instance against ergo. Delete:

- `src/stub-server.ts`
- `src/tools-only-stub.ts`
- the `bun stub` script
- the two README layout entries

Counter-argument worth airing: they're tiny, self-contained, useful as living examples for someone learning the MCP shape. Keeping them costs ~170 LOC.

cc @lead-pm — your call.

TeakBuilds:
Closes #68.

Deletes the two proof-of-concept stubs now that production `irc-server.ts` exists. Also removes the `bun stub` script and two README layout entries.

Verified no test or source file imports either stub (`rg -l` returns only `package.json` + `README.md`).

[worker-68]


AlexSc (approved):
Can we create a followup issue to review/refactor CLAUDE.md? beta milestone

PR #77, opened 2026-05-04
drop legacy chunker (#59)

### LINKED ISSUE #58: legacy chunker drops newlines on >max-lines messages
Surfaced from PR #53 / issue #18 ec5 work.

When an outbound message exceeds the negotiated `multilineMaxLines` (so the multiline batch path bails to legacy chunking), `sendLegacyChunks` uses `client.say()` which pre-splits on \`\\n\` before sending each line as an independent PRIVMSG. Receivers coalesce the messages back into a single notification but **without newline separators** — \`line0\\nline1\\nline2\` becomes \`line0line1line2\`.

This means roost agents cannot send messages that exceed ergo's max-lines limit (200) without silently corrupting them.

## Acceptance

- After legacy fallback, the round-trip text is byte-identical to what was sent (newlines preserved)
- Likely fix: have `sendLegacyChunks` send `\\n`-bearing wire chunks (PRIVMSG with embedded newline impossible per RFC; needs another mechanism — multiple PRIVMSGs with re-injected newline on receiver, or refuse and surface a clear error)

## Background

Found while writing test ec5 in #18. Initial test asserted text round-trip; switched to "tool doesn't error + first notification arrives" because the round-trip is currently lossy. ec5 carries a TODO comment pointing to this issue.

[lead-pm]

### LINKED ISSUE #59: drop legacy chunker once draft/multiline is ubiquitous
Now that ergo (our primary server) speaks draft/multiline cleanly, the legacy chunk-and-buffer path in sendLegacyChunks / recvBuffers is dead weight. It was added for ngircd compatibility (#8) but the multiline batch path is strictly better.

Removing it would let us delete sendLegacyChunks, the LEGACY_MARKER_RE strip, the recvBuffers map, INITIAL_BUFFER_MS, EXTENDED_BUFFER_MS, and the heuristic buffering in the PRIVMSG handler. Also resolves #58 (legacy path drops newlines) as a side effect.

Depends on confirming no production deployments still target a non-multiline server.

TeakBuilds:
closes #59, closes #58

drops sendLegacyChunks, recvBuffers, INITIAL/EXTENDED_BUFFER_MS, MAX_CHUNK_BODY, LEGACY_MARKER_RE, splitText, stripLegacyMarker. multiline is now unconditional. max-lines guard warns and proceeds instead of falling back.

~180 LOC gone. 18/18 tests green.

[worker-59]

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -537,8 +537,9 @@ export function createMcpServer(ircClient: any, config: McpServerConfig): { serv
       )
     } else {
       process.stderr.write(
-        `roost-irc[${NICK}]: draft/multiline NOT enabled (server caps: ${enabled.join(',') || '(none)'})\n`,
+        `roost-irc[${NICK}]: draft/multiline NOT enabled (server caps: ${enabled.join(',') || '(none)'}) — exiting, server must support draft/multiline\n`,
       )
+      process.exit(1)
     }
     process.stderr.write(
       enabled.includes(CAP_CHATHISTORY)
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -34,16 +34,12 @@ import IRC from 'irc-framework'
 import { MULTILINE_LINE_BYTES } from './constants.js'
 import {
   IrcMessage,
-  splitText,
   splitLineForMultiline,
   newBatchId,
-  stripLegacyMarker,
   reassembleMultilineBatch,
 } from './irc-lib.js'
 
 const SOURCE_NAME = 'roost-irc'
-const INITIAL_BUFFER_MS = 250
-const EXTENDED_BUFFER_MS = 2000
 const CAP_CHATHISTORY = 'chathistory'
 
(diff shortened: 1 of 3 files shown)


src/irc-server.ts
```diff
@@ -620,7 +537,7 @@ export function createMcpServer(ircClient: any, config: McpServerConfig): { serv
       )
     } else {
       process.stderr.write(
-        `roost-irc[${NICK}]: draft/multiline NOT enabled (server caps: ${enabled.join(',') || '(none)'}); falling back to legacy chunker\n`,
+        `roost-irc[${NICK}]: draft/multiline NOT enabled (server caps: ${enabled.join(',') || '(none)'})\n`,
```
AlexSc: We should probably fail to start the MCP in this case, since agents will immediately try sending multiple lines and only the first will go through.

AlexSc (changes requested):

AlexSc (approved):

PR #81, opened 2026-05-04
test(mcp-core): regression tests for waitForNotification waiter cleanup

### LINKED ISSUE #36: test helpers: timed-out waiters leak in waitForNotification
spotted by /simplify on #35.

the timeout handler does `waiters.findIndex(w => w.resolve === resolve)` but `w.resolve` is a wrapper closure, not the raw Promise `resolve` — so the comparison always misses and the waiter is never removed. dead entries pile up and get checked on every future notification.

harmless in short-lived tests, but worth a proper fix.

[worker-32]

TeakBuilds:
closes #36

the bug (raw `resolve` in findIndex vs wrapper stored in waiters) was already fixed in 517a93e when mcp-core.ts was created. this adds the regression guard that was missing.

**changes:**
- `McpHandle` gets `waiterCount: () => number` (exposes internal waiter state for testing)
- two new tests in `helpers.test.ts`: timeout path and success path both assert `waiterCount() === 0` after cleanup

[worker-36]

### DIFF
--- a/test/helpers.test.ts
+++ b/test/helpers.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, beforeAll } from 'bun:test'
 import { startErgo, isErgoAvailable, type ErgoContext } from './helpers/ergo.js'
 import { startMcp } from './helpers/mcp.js'
+import { startMcpInProcess } from './helpers/mcp-inprocess.js'
 import { connectPeer } from './helpers/peer.js'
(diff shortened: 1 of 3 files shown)


AlexSc:
Did this get an agent review?

AlexSc (approved):

AlexSc (approved):

Issue #83, opened 2026-05-04
tests: stabilize timing-flaky multiline/unread tests


TeakBuilds:
[lead-pm] Surfaced from worker-36 during #36 work — worker reported failure count varies between runs (5 → 9) on identical main commit.

Pre-existing, timing-sensitive, all in the multiline/unread surface. Blocking the value of #80 (CI) — fix this before turning CI on, otherwise CI will be noisy.

## Known-flaky tests (per worker-36 enumeration)

- **irc-server-inprocess.test.ts** — \`channel_message/channel_history/channel_ack clears unread\`
- **irc-server-inprocess.test.ts** — \`send reply includes unread nudge\`
- **outbound.test.ts** — \`channel_message long multiline reassembles to original\`
- **multiline-edge-cases.test.ts** — \`empty lines\`, \`trailing newline\`, \`byte boundary\`, \`max-lines timeout\`, \`mix of long+short\`

## Hypotheses

- Race between \`waitForNotification\` and the multiline batch reassembly arriving in fragments
- BATCH end / message dispatch timing on the receiver side
- Possibly some shared state across tests (notifications array carryover?)

## Why alpha

Required precondition for #80 (CI). Without this, CI signal will be noise. May also be incidentally improved by #65 (in-process migration); worth re-running after #65 merges to see what's left.

PR #86, opened 2026-05-04
tests: migrate subprocess MCP tests to in-process harness

### LINKED ISSUE #65: tests: migrate subprocess MCP tests to in-process InMemoryTransport harness
Surfaced from PR #61 approval.

PR #61 introduced an in-process MCP test harness (\`test/helpers/mcp-inprocess.ts\`) using the SDK's \`InMemoryTransport\`. It's faster, gives bun's coverage instrumentation visibility into \`src/irc-server.ts\`, and avoids per-test subprocess startup. Several existing tests still drive the MCP via \`StdioClientTransport\` against a spawned subprocess.

Migrate as many of those as cleanly fit onto the in-process harness. Keep a small smoke-coverage subprocess test for the actual binary path so the wiring isn't completely untested.

**Blocked on #63** (Opus full /simplify pass) — let that pass land first so we don't churn tests we're about to refactor.

Doesn't need Opus.

[lead-pm]

TeakBuilds:
Closes #65.

## Summary
- Creates in-process equivalents for inbound, outbound, chathistory, reconnect, and multiline (tests 1–4) using `InMemoryTransport` via `startMcpInProcess`
- Extends `startMcpInProcess` with optional config overrides (`joinHistoryLines`, `historySize`, `joinHistoryMinutes`)
- Slims subprocess files to only what can't run in-process: SIGUSR1/kill+restart lifecycle tests, and 2 timing-sensitive multiline boundary tests
- Deletes `inbound.test.ts`, `outbound.test.ts` (fully replaced)

## What stays subprocess and why
- `chathistory.test.ts`: SIGUSR1 test requires `process.kill` on subprocess pid
- `reconnect.test.ts`: kill+restart ergo test is inherently a subprocess lifecycle test
- `multiline-edge-cases.test.ts` (tests 5+6): "exceeding max-lines" and "mix of long+short lines" — these time out intermittently under the in-process harness; tracked in #83

## Note on pre-existing failures
4 tests in `irc-server-inprocess.test.ts` (`channel_message/history/ack clears unread`, `send reply includes unread nudge`) were already failing on main before this PR — channel names like `#ip-unread2` contain the string "unread", tripping the `.not.toContain('unread')` assertion. Out of scope here.


AlexSc (changes requested):
Is there any reason the in-process tests need to be in different suites? Or if they do, can we split out the not-in-process ones? Mostly I'm looking to minimize the diff and make it easier to keep track of what the expected happy path (more in-process tests so we get coverage) is.

Also has anyone checked if this improves our coverage story?

AlexSc (approved):
Note there are merge conflicts that need to be dealt with

PR #89, opened 2026-05-04
feat(service): roost service start|status|stop + watcher prompt

### LINKED ISSUE #67: roost: local lifecycle story for project background services (dispatcher etc.)
Bringing up the lead-pm today required a manual checklist: check tmux, check ps, check config, spawn daemon in tmux window, tail log to confirm. That's friction we hit every fresh session, and it'll only get worse as projects accumulate more side processes (mock backends, log tailers, browser automation).

Roost shouldn't bake in what those processes *are* — that's per-project. But Roost is a natural place to own the **convention** for declaring and supervising them.

## Shape

- Project declares services in `.roost/services.json` (or a section of an existing roost config): `name`, `cmd`, `cwd`, optional `healthcheck`
- `bin/roost service ensure <name>` is idempotent: if alive, no-op; else launch into a tmux window (`roost-services:<name>`) and confirm
- `bin/roost service status` lists all declared services with up/down state
- `bin/roost service stop <name>` for clean shutdown
- Lead-pm-style spawn loops `ensure` over declared services on boot — one line in the prompt instead of a checklist

## Out of scope

- What any specific service does. The dispatcher / orchestrator example lives in #5 (beta).
- Restart policies more complex than \"start once, leave running\" — keep it dumb for alpha.

## Why alpha

Every fresh lead-pm session pays the friction. Sample dispatcher (#5) can assume this exists.

[lead-pm]

TeakBuilds:
Closes #67.

## What's in here

- `roost service start|status|stop` — dumb tmux subprocess manager. Spawns into `roost-services:<basename>`. `start` is idempotent (window exists + live process). `stop` on a missing service is a friendly no-op.
- `prompts/watcher.md` — canonical watcher prompt (v2 from issue #67 comment 3), extended with on-boot `roost service start bin/orchestrator_poll`.
- `start.md` — replaces the manual tmux checklist with a single `roost spawn watcher --prompt-file watcher` line.
- `roost spawn --prompt-file watcher` bare-name resolution: no path separator → `${ROOST_DIR}/prompts/watcher.md`.

## Tests

`test/service.test.ts` — skips if `tmux` not installed (matches ergo pattern). Covers start, idempotent start, status, stop, stop-when-missing.

[worker-67]

### DIFF
--- a/.claude/commands/watcher.md
+++ b/.claude/commands/watcher.md
@@ -1,6 +1,10 @@
-You are `watcher`, a small haiku agent on the local Roost (ergo on 127.0.0.1:6667). You are joined to #leads-roost-dev. The lead PM is @lead-pm and the human is @alex.
+---
+description: Roost watcher haiku — supervises the orchestrator and mutates the watch list in response to channel/DM commands.
+argument-hint: [lead-nick] [human-nick]
+---
+You are `watcher`, a small haiku agent on the local Roost (ergo on 127.0.0.1:6667). You are joined to #leads-roost-dev. The lead PM is @$0 and the human is @$1.
 
-You exist as a throwaway scaffold for issue #67. Your single responsibility: listen for watch-list commands in #leads-roost-dev or as DMs from @lead-pm or @alex, and mutate `/Users/alex/Dev/GoCarrot/roost/.orchestrator/config.json` accordingly. The orchestrator daemon re-reads config each tick (~60s).
+You exist as a throwaway scaffold. Your single responsibility: listen for watch-list commands in #leads-roost-dev or as DMs from @$0 or @$1, and mutate `.orchestrator/config.json` (relative to your working directory) accordingly. The orchestrator daemon re-reads config each tick (~60s).
 
 ## IMPORTANT — tooling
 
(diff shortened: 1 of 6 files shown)


AlexSc (changes requested):
Couple of problems:
- In the plugin, I think roost dir resolves to the plugin cache directory, which humans are unlikely to view and shouldn't edit. This means the bare prompt file pulls whatever we shipped instead of what the user might be trying to do locally. Confusing and undesirable.
- Related, the prompt is extremely specific to our setup. Hardcoded user names, etc.

I think we want to prepare the example prompt as a templated command, using $ARGUMEMTS et al (search the Claude code docs), cut the bare prompt file path, and locally symlink our example prompts to Claude code commands so we can start the watcher with a slash command.

Thoughts?

AlexSc:
I'm not sure a command to spawn the watcher gets us much. I was thinking a command the watcher is prompted to run at launch, so the lead runs something like roost spawn watcher -c '#leads' --prompt '/watch-orchestrator lead alex', and watch-orchestrator is the templated version of our current prompt, symlinked to project commands.

Does that make sense?

AlexSc:
Yep, you got it. Integrators choice on the name.

AlexSc (approved):

PR #101, opened 2026-05-04
docs: clarify channel_ack semantics

### LINKED ISSUE #99: docs: clarify channel_ack semantics — sending a message implies ack
## Doc gap

Agents (lead-pm specifically observed today) are double-acking channels: sending a message and then immediately calling \`channel_ack\` on the same channel. That's wasteful — sending a message should mark the channel read.

## Fix

Update wherever the MCP \`channel_ack\` semantics are documented (the tool description string, plus any agent-facing docs in \`docs/\` or skill files) to say:

> \`channel_ack\` is only needed when you read a channel's traffic but have nothing to say in response. Posting any message to a channel (channel_message / direct_message) implicitly acks it. Don't ack a channel you just messaged.

If the implementation doesn't actually auto-ack on send, that's a separate bug — but the docs should reflect what the right mental model is.

## Why

Reduces redundant tool calls and operator confusion about when to ack.

TeakBuilds:
closes #99

Updated `channel_ack` tool description and watcher skill to say: posting any message implicitly acks the channel — only call `channel_ack` when reading with nothing to say in response.

[worker-99] two-line change, docs only

### DIFF
--- a/.claude/commands/watcher.md
+++ b/.claude/commands/watcher.md
@@ -12,7 +12,7 @@ You have IRC tools as MCP. Use them. Do NOT use Bash/nc/raw IRC protocol.
 
 - Send to channel: `channel_message`
 - Send DM: `direct_message`
-- Read: `channel_history`, `channel_ack`
+- Read: `channel_history`, `channel_ack` (only when you read but have nothing to say — sending a message implicitly acks the channel)
 
 ## Commands you accept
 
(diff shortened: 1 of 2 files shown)


TeakBuilds (commented):
[reviewer-99] two small wording notes on the irc-server.ts description:

1. "do not call channel_ack on a channel you just messaged" is redundant — "implicitly acks it" already covers this. Drop the second clause.
2. "channel's traffic" reads oddly; "channel's messages" or just "a channel" is cleaner.

watcher.md parenthetical is fine as-is.

AlexSc (approved):

PR #105, opened 2026-05-04
fix: thread permbot deny reason into PermissionRequest message

### LINKED ISSUE #97: permbot: deny reason not propagated to agent
## Bug

`bin/irc-permission-prompt`'s `emit()` writes only `{behavior: decision}` for both allow and deny — no `message` field. So when the operator denies via permbot with rationale (e.g. \`n - stop polling, GitHub is down\`), the agent gets a bare deny and never sees the why. It just retries with a different angle.

Hit live today during the GH outage: lead-pm denied worker-80's CI-poll requests with reasons, worker-80 kept retrying because the rationale never reached the agent.

## Fix shape

In `emit()`, when `decision == \"deny\"` (or also \"allow\"), include the operator's reply as `decision.message`. One-line change in the JSON shape.

```python
out = {\"hookSpecificOutput\": {\"hookEventName\": \"PermissionRequest\", \"decision\": {\"behavior\": decision}}}
if reason:
    out[\"hookSpecificOutput\"][\"decision\"][\"message\"] = reason
```

Also need to thread the operator's full reply (not just \`y\`/\`n\`) into the deny path. Today permbot strips the first token to decide allow/deny but the rest of the reply is dropped — preserve it as the message.

## Why this matters

Lead-pm uses denies to redirect workers (\"stop, read the local file instead of web searching\"). When the rationale doesn't reach the worker, lead-pm has to escalate to shutdown or \`roost send\` to inject a TUI message — both heavier than necessary.

TeakBuilds:
Fixes #97.

`emit()` was writing `{behavior: decision}` only — no `message` field. Operator rationale was silently dropped and the agent got a bare deny with no context.

**Changes** (single file, `bin/irc-permission-prompt`):
- split operator reply on first whitespace: first token = decision, rest = reason verbatim
- include reason as `decision.message` for both allow and deny paths

**Live verification**: denied a Bash call with "n stop polling, use the cached file instead", confirmed the worker's next response referenced that exact phrase (not a paraphrase).

[worker-97]

### DIFF
--- a/bin/irc-permission-prompt
+++ b/bin/irc-permission-prompt
@@ -34,16 +34,19 @@ SOCKET_SAFETY_TIMEOUT = 570  # just under Claude Code's 600s hook default
 PASSTHROUGH_PREFIXES = ("mcp__roost-irc__", "mcp__plugin_roost_roost-irc__")
 
 
-def emit(decision: str, reason: str) -> None:
+def emit(decision: str, reason: str = "") -> None:
     if decision == "ask":
         # No decision → Claude Code shows the terminal permission prompt
         sys.stderr.write(f"perm-hook: deferring to terminal ({reason})\n")
         sys.exit(0)
+    dec: dict = {"behavior": decision}
+    if reason:
+        dec["message"] = reason
     json.dump(
         {
             "hookSpecificOutput": {
                 "hookEventName": "PermissionRequest",
-                "decision": {"behavior": decision},
+                "decision": dec,
             }
         },
         sys.stdout,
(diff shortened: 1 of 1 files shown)


TeakBuilds (commented):
[reviewer-97] Clean diff, no issues. A few notes:

- `dec` construction with the conditional `message` field is the right pattern — no footguns.
- `parts = reply.strip().split(maxsplit=1) if reply.strip() else []` correctly handles whitespace-only replies.
- The allow path no longer emits a hardcoded `"approved via IRC"` message when the operator gives no reason — intentional improvement, operator silence stays silent.
- The deny path loses the old `denied via IRC: {reply!r}` debug string, which is fine; that was noise, not rationale.

LGTM.

AlexSc (approved):

PR #107, opened 2026-05-04
fix: channel_leave waits for own PART ack

### LINKED ISSUE #100: channel_leave should wait for own PART ack (mirror channel_join)
## Problem

`channel_leave` is fire-and-forget — it calls `ircClient.part(channel)` and returns "parted" immediately, before the IRC server has acknowledged the PART. This is inconsistent with `channel_join`, which waits via `join_resolvers` for the server to confirm.

## Why it matters

Surfaced via the CI race in #93 — callers that observe state right after `channel_leave` returns (here: the test peer waiting for PART) can race with the actual server-side part. The PR worked around this by reordering the test, but the asymmetric API will keep biting future callers.

## Proposed fix

Mirror `channel_join`'s pattern: add `part_resolvers: Map<string, Array<(ok: boolean) => void>>`, await it in the `channel_leave` tool handler, resolve from the `'part'` event handler when `event.nick === NICK`. Use the same 5s timeout.

## Scope notes

Behavior change — callers will see `channel_leave` block until ack (or 5s timeout). Existing tests that just `await callTool(channel_leave)` will continue to work (they ignore the return value or expect success).

Light lift — ~10 lines in `src/irc-server.ts`.

TeakBuilds:
closes #100

Mirrors `channel_join`'s `join_resolvers` pattern — adds `part_resolvers`, awaits the `'part'` event for our own nick before returning (5 s timeout), resolves before `channelUsers.delete` to avoid re-join races.

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -64,6 +64,7 @@ export function createMcpServer(ircClient: any, config: McpServerConfig): { serv
   let irc_ready = false
   let hasRegistered = false
   const join_resolvers = new Map<string, Array<(ok: boolean) => void>>()
+  const part_resolvers = new Map<string, Array<(ok: boolean) => void>>()
 
   // Max lines per draft/multiline batch, from the cap value
   // (`draft/multiline=max-bytes=16384,max-lines=200`). Pre-negotiation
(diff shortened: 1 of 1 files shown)


TeakBuilds (commented):
[reviewer-100] clean mirror of join_resolvers, no simplifications needed. one nit: `channel_join` lowercases the channel at line 441 (`.toLowerCase()`) but `channel_leave` doesn't. If a caller passes `#General`, the resolver is keyed as `#General` but the IRC server's PART event likely comes back as `#general`, so the lookup misses and the call times out. Suggest adding `.toLowerCase()` to match the join pattern.

AlexSc (approved):

PR #108, opened 2026-05-04
feat: channel_list authoritative via WHOIS

### LINKED ISSUE #98: MCP: channel_list should optionally consult authoritative server state
## Suggestion

Today \`channel_list\` is served exclusively from the local in-memory cache. After #92 lands we'll have cache invalidation on reconnect + auto-rejoin, but there's still a window where the cache can drift if the server-side membership changes for reasons we don't track (KICK by oper, channel deletion, etc.).

Add an optional \`authoritative=true\` flag to \`channel_list\` that issues a fresh \`LIST\` (or \`WHOIS\` on self) to the server and reconciles the cache before returning. Default stays as-is (fast cache read) so callers that don't care don't pay the round-trip.

## Why

Supervisor agents (lead-pm) use \`channel_list\` to decide where to relay messages. A drift-induced wrong answer manifests as silently missed activity. Having an explicit \"verify with server\" option lets supervisors do a quick consistency check at strategic points (e.g. after long idles, after reconnect events) without paying the cost on every call.

## Note

Discovered live during the silent-reconnect incident on 2026-05-04 — surfaced in conversation but not previously filed.

TeakBuilds:
closes #98

`channel_list` now calls `ircClient.whois(NICK)` and parses the 319 RPL_WHOISCHANNELS response instead of reading the local `channelUsers` cache. 5s timeout → `isError: true`, same pattern as `channel_join`. Unread overlay unchanged.

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -379,7 +379,7 @@ export function createMcpServer(ircClient: any, config: McpServerConfig): { serv
       },
       {
         name: 'channel_list',
-        description: 'List all channels currently joined by this MCP instance. Served from a local cache (no network round-trip); the cache is kept current via JOIN/PART/KICK/QUIT events.',
+        description: 'List all channels currently joined by this MCP instance. Issues a live WHOIS query to the IRC server on every call for an authoritative result.',
         inputSchema: {
           type: 'object',
           properties: {},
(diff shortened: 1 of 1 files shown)


TeakBuilds (commented):
[reviewer-98] A few things worth looking at:

**Scope drift**: issue #98 asked for an optional `authoritative=true` flag with the cache as default. This PR replaces the cache path entirely — every call now pays the WHOIS round-trip. That's a simpler API (no flag to explain), but it's a design call worth making explicitly. Is the 5s worst-case latency on every call acceptable for all callers (e.g. supervisors checking after idle)?

**`&` in mode-prefix regex**: `.replace(/^[@+%&~]+/, '')` strips `&` as a mode prefix. `&` is also a valid channel type prefix (e.g. `&localchannel`). On ergo this is probably never triggered, but the regex would corrupt `&`-prefixed channel names. If ergo only surfaces `#` channels in WHOIS 319, fine — just worth confirming.

**Sentinel inconsistency**: `channel_join` uses `false` as the timeout sentinel; `channel_list` uses `null`. Both work, but aligning them (or extracting the 5000ms constant shared between them) would make the pattern easier to scan.

Nothing here is a blocker, but the scope question is worth a comment in the PR from the author confirming the always-WHOIS choice was intentional.

AlexSc:
Do we have tests?

AlexSc (approved):

PR #109, opened 2026-05-04
fix: pass all args through roost service start to tmux

### LINKED ISSUE #106: watcher: `roost service start bin/orchestrator_poll` doesn't actually run the daemon
## Bug

The watcher skill's boot-time instruction is:
```
roost service start bin/orchestrator_poll
```

`roost service start` takes only the first arg as the command (`local cmd="$1"`), so any additional flags are dropped. `bin/orchestrator_poll` without args does a one-shot tick (`prints []`) and exits immediately.

Result: `roost service status` reports the service "started" briefly, then it's gone. `dispatcher-roost` never appears in any IRC channel. Watcher updates the watch list but no events route, because the dispatcher isn't running.

## Repro

1. Spawn watcher per docs
2. Approve its `roost service start bin/orchestrator_poll` permission request
3. `bin/roost service status` → `(no services)`
4. `tmux ls` → no `roost-services` session

## Fix shape

Pick one:

1. **Make `roost service start` pass remaining args** as the command — `tmux new-window ... "$@"` — and update the watcher skill to say `bin/orchestrator_poll --daemon`.
2. **Make orchestrator_poll daemon-by-default** when run as a service (e.g. detect ROOST_SERVICE env, or just default to daemon and require explicit `--once` for the one-shot mode).
3. **Wrapper script** `bin/orchestrator_poll_daemon` that just execs `orchestrator_poll --daemon`.

(1) is cleanest IMO — the service runner is the one with the obscure constraint. Filed by [lead-pm] after observing the dispatcher silently absent for a full Wave 1 of work.

TeakBuilds:
Fixes #106.

`cmd_service start` captured only `$1` and passed a single-word string to tmux, silently dropping any trailing flags. Changed to use `$@` so `roost service start bin/orchestrator_poll --daemon` works as expected. Also updated the watcher skill's boot instruction to include `--daemon`.

Smoke-tested: `roost service start /bin/sleep 600` → `ps` confirmed `sleep 600` in the tmux pane → status showed `up` → stopped clean.

### DIFF
--- a/.claude/commands/watcher.md
+++ b/.claude/commands/watcher.md
@@ -29,7 +29,7 @@ A single message may contain multiple commands, separated by newlines, semicolon
 
 ## Behavior rules
 
-- **On boot, first action:** run `roost service start bin/orchestrator_poll` to ensure the dispatcher daemon is up.
+- **On boot, first action:** run `roost service start bin/orchestrator_poll --daemon` to ensure the dispatcher daemon is up.
 - Read the config file before each mutation (don't cache).
 - Pretty-print JSON on write (2-space indent, trailing newline).
 - If you don't recognize a command, ignore it silently.
--- a/bin/roost
+++ b/bin/roost
@@ -489,9 +489,8 @@ cmd_service() {
         echo "error: service start requires a command" >&2
         exit 1
       fi
-      local cmd="$1"
       local name
-      name="$(basename "${cmd}")"
+      name="$(basename "$1")"
       # Idempotency: window exists and process is alive → no-op
       if tmux list-windows -t "${SERVICE_SESSION}" -F '#{window_name}' 2>/dev/null \
            | grep -qx "${name}" && _service_is_alive "${name}"; then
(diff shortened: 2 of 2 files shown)


TeakBuilds (commented):
[reviewer-106] looks good, no simplifications to suggest.

the fix is minimal and correct: drop the `local cmd` alias (one fewer variable) and pass `"$@"` instead of `"${cmd}"` so trailing args survive to tmux.

a few things i checked:

- **path with spaces in `$1`**: `basename "$1"` handles it fine since `$1` is quoted.
- **`"$@"` in tmux**: tmux exec's the argv directly, so each element lands as a separate arg — correct behavior, spaces in individual args preserved.
- **window name matching**: `grep -qx "${name}"` uses the basename so the idempotency check is unaffected.
- **watcher.md**: doc change reads correctly — `--daemon` is now part of the boot command so watcher agents get the right incantation.

nothing to cut or rewrite here.

AlexSc:
Did you run shellcheck? And can you check if shellcheck is running in CI? if it isn't can we get a followup alpha issue to get shellcheck in CI?

AlexSc (approved):

PR #139, opened 2026-05-06
maint: delete dead bindings in roost-permbot

### LINKED ISSUE #130: chore: delete dead state in roost-permbot (P1)
Three dead bindings in `bin/roost-permbot`:

- `dlog` early-return guard (line 57-58): `DEBUG_LOG` is always truthy (set via env or `os.path.join` fallback), so `if not DEBUG_LOG: return` never fires. Delete the guard.
- `IRC.fileno` method (line 75-76): `sel.register` registers `irc.sock` directly (line 135), never calls `irc.fileno()`. Delete the method.
- `IRC.registered` flag (line 73 + 255): set in `__init__` and on `001`, never read anywhere. Delete both assignments.

**Acceptance:**
- All three dead bindings removed, no functional change, tests pass

TeakBuilds:
Closes #130

Three dead bindings removed, no functional change:

- `dlog` guard: `if not DEBUG_LOG: return` — unreachable, DEBUG_LOG always truthy
- `IRC.fileno` — never called, sel.register uses irc.sock directly
- `IRC.registered` flag — set but never read

ast parses clean, grep finds no other references to removed names, bun test 27/27 pass.

### DIFF
--- a/bin/roost-permbot
+++ b/bin/roost-permbot
@@ -54,8 +54,6 @@ DEBUG_LOG = os.environ.get("ROOST_PERM_DEBUG_LOG") or os.path.join(os.path.dirna
 
 
 def dlog(msg: str) -> None:
-    if not DEBUG_LOG:
-        return
     try:
         with open(DEBUG_LOG, "a") as f:
             f.write(f"[{time.strftime('%H:%M:%S')} {NICK}] {msg}\n")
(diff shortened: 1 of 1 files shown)


TeakBuilds (commented):
[reviewer-139] LGTM, simplify clean — no reuse, quality, or efficiency issues. Pure deletion, nothing to flag.

AlexSc (approved):

PR #141, opened 2026-05-06
fix: DM operator on terminal fallback in irc-permission-prompt

### LINKED ISSUE #132: fix: terminal fallback in irc-permission-prompt is invisible to remote operator (PP1)
When `irc-permission-prompt` can't reach the permbot daemon, it falls back to Claude Code's local terminal prompt. For an unattended worker (the common case), there's no human watching the terminal — the worker blocks indefinitely on a prompt nobody sees.

Fix: when falling back to terminal, also DM the operator via `ROOST_PERM_TARGET` with the tool summary and reason. Doesn't override the terminal prompt; just makes the block visible to whoever spawned the worker.

**Acceptance:**
- Terminal fallback sends a DM to `ROOST_PERM_TARGET` (if set) with the reason before deferring
- Interactive sessions unaffected (DM is a notification, not a gate)

TeakBuilds:
Closes #132

When the permbot socket is unreachable (or returns an unrecognized reply), the hook deferred to Claude Code's terminal prompt silently — invisible to an unattended worker's operator. Now fires a DM to `ROOST_PERM_TARGET` (if set) with the tool summary and reason before deferring.

**Changes:**
- `bin/irc-permission-prompt`: `send_fallback_dm()` — opens a short-lived IRC connection, registers as `pnotify-{WORKER}`, PRIVMSGs the summary+reason to `ROOST_PERM_TARGET`, QUITs. 10s timeout. Called before both fallback `emit("ask", ...)` sites.
- `bin/roost`: passes `ROOST_PERM_TARGET`, `ROOST_PERM_HOST`, `ROOST_PERM_PORT` into the worker tmux env alongside `ROOST_PERM_SOCK`.
- `test/irc-permission-prompt.test.ts`: Bun TCP stub verifies DM fires with the right content; second test verifies no connection when `ROOST_PERM_TARGET` is unset.

[worker-132]

### DIFF
--- a/bin/irc-permission-prompt
+++ b/bin/irc-permission-prompt
@@ -29,6 +29,9 @@ import time
 
 WORKER = os.environ.get("ROOST_IRC_NICK", "unknown")
 SOCK_PATH = os.environ.get("ROOST_PERM_SOCK", "")
+PERM_TARGET = os.environ.get("ROOST_PERM_TARGET", "")
+PERM_HOST = os.environ.get("ROOST_PERM_HOST", "127.0.0.1")
+PERM_PORT = int(os.environ.get("ROOST_PERM_PORT", "6667"))
 SOCKET_SAFETY_TIMEOUT = 570  # just under Claude Code's 600s hook default
 
 PASSTHROUGH_PREFIXES = ("mcp__roost-irc__", "mcp__plugin_roost_roost-irc__")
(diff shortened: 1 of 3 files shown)


TeakBuilds (commented):
[reviewer-141] overall clean. a few things worth knowing:

**nick collision / stale registration** — if a prior `pnotify-{WORKER}` invocation crashed before QUIT (killed process, SIGKILL, etc.), ergo holds the nick until its ping-timeout fires. Next fallback DM connects fine but never gets 001 (server sends 433 instead), so the registration loop spins until the 10s wall-clock deadline. That's a 10s drag on the terminal fallback. Not catastrophic — bounded, stderr-logged, fallback still fires — but worth being aware of. No 433 branch exists to bail early.

**ROOST_PERM_HOST hardcoded in bin/roost** — line 308 always passes `ROOST_PERM_HOST=127.0.0.1`; if someone ever points ergo at a non-loopback address this silently ignores it. Probably intentional for now.

**QUIT fires unconditionally** (line 219, outside `if registered:`) — good, tells ergo to drop the half-registered conn cleanly regardless of outcome.

**control flow at line 310** — safe. Both `emit("allow")` and `emit("deny")` call `sys.exit(0)`, so `send_fallback_dm` only runs if neither matched.

**test coverage** — happy path + no-target-skips-connect covered. The 433/stale-nick path (registered=False timeout) is not tested, but it's also the safe-degradation path so probably fine.

**perm_target empty string** — Python guard (`if not PERM_TARGET: return`) handles the case where `--perm-target` wasn't passed and `ROOST_PERM_TARGET=` arrives as empty string. Fine.

Nothing blocking. The stale-nick 10s lag is the one thing worth a follow-up ticket if it shows up in practice.

AlexSc:
lead-pm: Can we create a beta issue to address the hardcoded loopback address? I think it shows up in a couple of places

AlexSc (approved):

Issue #135, opened 2026-05-06
feat: document and normalize unread suffix contract (L13)


TeakBuilds:
The unread suffix appears after `channel_message`, `direct_message`, and `channel_list` — but not after `channel_ack`. Agents learn the suffix exists by observation; zero tool schema descriptions mention it. The natural follow-on to acking a channel ("I just finished reading #foo, what else needs me?") gets no unread nudge today.

**Code change:** add unread suffix emission to `channel_ack` (matching its siblings). One-line addition after `unread.delete(channel)`.

**Doc change:** update all 4 tool schema descriptions to document the suffix: "Response includes a trailing 'unread:' summary listing other channels with pending messages." Update the MCP `instructions` string to describe the suffix as a passive nudge alongside the existing SIGUSR2 mention.

**Out of scope (separate issues):** suffix on `channel_history`; mention-flag dimension; unread pruning/decay.

**Acceptance:**
- `channel_ack` emits the unread suffix after clearing the channel
- `channel_message`, `direct_message`, `channel_list`, `channel_ack` tool descriptions each document the suffix
- MCP `instructions` string describes the suffix
- Tests cover the suffix emission from `channel_ack`

PR #142, opened 2026-05-06
feat: RoostIrcClient interface (R1)

### LINKED ISSUE #118: refactor: define RoostIrcClient interface (R1)
`createMcpServer` is a 780-line closure because there's no typed seam between the IRC client layer and the MCP layer. Before extracting that seam, we need to define what it looks like.

Create `src/irc-client.ts` with the `RoostIrcClient` interface and supporting types (`MessageMeta`, `MembershipExtras`, `ConnectOpts`, `UnreadInfo`). Types only — no implementation, no runtime change.

```ts
export interface RoostIrcClient {
  connect(opts: ConnectOpts): void
  isReady(): boolean

  join(channel: string, force?: boolean): Promise<boolean>
  leave(channel: string): Promise<boolean>
  say(target: string, text: string): { chunks: number; mode: 'single' | 'multiline' }
  whoisChannels(nick?: string): Promise<string[] | false>

  getHistory(key: string, limit?: number): IrcMessage[]
  getUsers(channel: string): string[]
  getUnread(): Map<string, UnreadInfo>
  ackUnread(key: string): void

  clearDedupeCache(): void

  on(event: 'message',    handler: (msg: IrcMessage, meta: MessageMeta) => void): void
  on(event: 'membership', handler: (kind: 'join'|'leave'|'nick', nick: string, channel: string, extras: MembershipExtras) => void): void
  on(event: 'system',     handler: (kind: 'disconnected'|'reconnected', content: string) => void): void
}
```

Gives R2–R5 a compiler-enforced contract from day one. Without it, parallel slices drift silently.

**Acceptance:**
- `src/irc-client.ts` compiles with the interface matching the shape above
- Supporting types (`MessageMeta`, `MembershipExtras`, `ConnectOpts`, `UnreadInfo`) defined in the same file
- Tests pass, zero runtime change

TeakBuilds:
Closes #118

Types-only seam between the IRC client layer and the MCP layer. No runtime change.

- `RoostIrcClient` interface with `connect`, `isReady`, `join`, `leave`, `say`, `whoisChannels`, `getHistory`, `getUsers`, `getUnread`, `ackUnread`, `clearDedupeCache`, and three `on()` overloads for `message` / `membership` / `system` events
- Supporting types: `UnreadInfo` (promoted from local in createMcpServer), `MessageMeta`, `MembershipExtras`, `ConnectOpts`
- `IrcMessage` re-exported from irc-lib.ts

**Deviates from spec on `whoisChannels`:** issue #118 had `whoisChannels(nick?: string)` but reviewer-142 (TeakBuilds, issue author) flagged the optional param as dead surface area — nothing ever whois's a nick other than self. Dropped to `whoisChannels(): Promise<string[] | false>` per their review feedback.

**`ConnectOpts` is camelCase** (`autoReconnect`/`autoReconnectMaxRetries`); R2 translates to irc-framework's snake_case internally.

**`getUnread()` returns `ReadonlyMap`** to prevent callers from corrupting client state.

R2/R3 will implement this interface wrapping irc-framework. R4+ consume it. R2 absorbs `unread` + `seenFingerprints` state (currently in `createMcpServer`) including the `if (!historical)` unread guard.

### DIFF
--- a/src/irc-client.ts
+++ b/src/irc-client.ts
@@ -23,14 +23,15 @@ export interface MembershipExtras {
   newNick?: string
 }
 
+// camelCase — R2 translates to irc-framework's snake_case fields internally.
 export interface ConnectOpts {
   host: string
   port: number
   nick: string
   username?: string
   gecos?: string
-  auto_reconnect?: boolean
-  auto_reconnect_max_retries?: number
+  autoReconnect?: boolean
+  autoReconnectMaxRetries?: number
 }
 
 export interface RoostIrcClient {
@@ -40,11 +41,11 @@ export interface RoostIrcClient {
   join(channel: string, force?: boolean): Promise<boolean>
   leave(channel: string): Promise<boolean>
   say(target: string, text: string): { chunks: number; mode: 'single' | 'multiline' }
-  whoisChannels(nick?: string): Promise<string[] | false>
+  whoisChannels(): Promise<string[] | false>
 
   getHistory(key: string, limit?: number): IrcMessage[]
   getUsers(channel: string): string[]
-  getUnread(): Map<string, UnreadInfo>
+  getUnread(): ReadonlyMap<string, UnreadInfo>
   ackUnread(key: string): void
 
   clearDedupeCache(): void
--- a/src/irc-client.ts
+++ b/src/irc-client.ts
@@ -23,7 +23,6 @@ export interface MembershipExtras {
   newNick?: string
 }
 
-// camelCase — R2 translates to irc-framework's snake_case fields internally.
 export interface ConnectOpts {
   host: string
   port: number
--- a/src/irc-client.ts
+++ b/src/irc-client.ts
@@ -34,16 +34,25 @@ export interface ConnectOpts {
 }
 
 export interface RoostIrcClient {
+  // Fire-and-forget: MCP starts serving before IRC connects (returns isError until isReady()).
+  // Use isReady() + on('system') to track connection state.
   connect(opts: ConnectOpts): void
   isReady(): boolean
 
+  // force bypasses the already-joined cache; use to recover wedged state without restarting.
   join(channel: string, force?: boolean): Promise<boolean>
   leave(channel: string): Promise<boolean>
+  // Synchronous socket write — no protocol-level delivery ack for PRIVMSG.
   say(target: string, text: string): { chunks: number; mode: 'single' | 'multiline' }
   whoisChannels(): Promise<string[] | false>
 
+  // Served from local cache — no network round-trip. Note: on a freshly-joined channel
+  // these lag the join ack; NAMES (getUsers) and chathistory (getHistory) arrive via
+  // events after join() resolves.
   getHistory(key: string, limit?: number): IrcMessage[]
   getUsers(channel: string): string[]
+  // R2 increments on every non-historical inbound message; MCP reads this to build
+  // the unread suffix on tool responses.
   getUnread(): ReadonlyMap<string, UnreadInfo>
   ackUnread(key: string): void
 
--- a/src/irc-client.ts
+++ b/src/irc-client.ts
@@ -1,5 +1,4 @@
-// Types-only seam between the IRC client layer and the MCP layer (R1).
-// No runtime code — implementation lives in R2.
+// Pure types — typed seam between the IRC client layer and the MCP layer.
 
 import type { IrcMessage } from './irc-lib.js'
 export type { IrcMessage }
@@ -51,7 +50,7 @@ export interface RoostIrcClient {
   // events after join() resolves.
   getHistory(key: string, limit?: number): IrcMessage[]
   getUsers(channel: string): string[]
-  // incremented on each inbound non-historical message; reset by ackUnread.
+  // Incremented on every non-historical inbound message; tool handlers read this to build the unread suffix.
   getUnread(): ReadonlyMap<string, UnreadInfo>
   ackUnread(key: string): void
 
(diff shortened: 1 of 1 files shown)


TeakBuilds (commented):
[reviewer-142] types-only review — CI green, 3 findings:

**1. `getUnread(): Map<string, UnreadInfo>` returns a mutable live reference.** Caller can corrupt client state. Should be `ReadonlyMap<string, UnreadInfo>` or return a snapshot. Straightforward fix.

**2. `whoisChannels(nick?: string)` — the optional nick is dead surface area.** Current impl always passes self-nick; nothing ever whois's another user via this method. Either drop the param (`whoisChannels(): Promise<string[] | false>`) or require it and make the caller pass the nick explicitly. Optional-with-a-hidden-default is a trap for R4+ consumers.

**3. `ConnectOpts` mixes snake_case and camelCase.** `auto_reconnect` / `auto_reconnect_max_retries` are irc-framework field names leaked into the public interface. R2 could accept `autoReconnect`/`autoReconnectMaxRetries` and translate internally — keeps the seam clean. If you're intentionally mirroring irc-framework's shape, a comment saying so would help.

**Design consequence to confirm (not blocking):** `getUnread`, `ackUnread`, and `clearDedupeCache` on the interface mean R2 must absorb the `unread` map and `seenFingerprints` state currently living in `createMcpServer`. Looks intentional given the refactor plan, just making sure R2 knows to own historical-message unread exclusion (`if (!extras.historical)` guard at irc-server.ts:226).

Everything else is clean — MessageMeta/MembershipExtras match exactly what's used, on() overloads correctly coalesce the three raw event streams, emitUnreadSummary is rightly absent.

src/irc-client.ts
```diff
@@ -0,0 +1,56 @@
...
+  historical?: boolean
+}
+
+// Extras on membership events from join/part/kick/quit/nick.
+export interface MembershipExtras {
+  reason?: string
+  newNick?: string
+}
+
+// camelCase — R2 translates to irc-framework's snake_case fields internally.
```
AlexSc: Cut non-timeless comment

src/irc-client.ts
```diff
@@ -0,0 +1,56 @@
...
+  port: number
+  nick: string
+  username?: string
+  gecos?: string
+  autoReconnect?: boolean
+  autoReconnectMaxRetries?: number
+}
+
+export interface RoostIrcClient {
+  connect(opts: ConnectOpts): void
```
AlexSc, line 38: Should connect return a promise that resolves when connected?

src/irc-client.ts
```diff
@@ -0,0 +1,55 @@
...
+  isReady(): boolean
+
+  join(channel: string, force?: boolean): Promise<boolean>
```
AlexSc, line 42: What does force do? I feel like that deserves a comment.

src/irc-client.ts
```diff
@@ -0,0 +1,55 @@
...
+  leave(channel: string): Promise<boolean>
+  say(target: string, text: string): { chunks: number; mode: 'single' | 'multiline' }
```
AlexSc, line 45: Should this return a promise that resolves when the message is sent?

src/irc-client.ts
```diff
@@ -0,0 +1,55 @@
...
+  whoisChannels(): Promise<string[] | false>
+
+  getHistory(key: string, limit?: number): IrcMessage[]
+  getUsers(channel: string): string[]
+  getUnread(): ReadonlyMap<string, UnreadInfo>
```
AlexSc, line 54: Is the plan for the MCP populating the unread suffix (see #135) to have it call getUnread() as part of the tool calls/event receipts?

src/irc-client.ts
AlexSc: Also should we have the promise resolve with channel information (current users, history, etc), or handle that a layer up?

src/irc-client.ts
```diff
@@ -0,0 +1,55 @@
...
+  connect(opts: ConnectOpts): void
+  isReady(): boolean
+
+  join(channel: string, force?: boolean): Promise<boolean>
+  leave(channel: string): Promise<boolean>
+  say(target: string, text: string): { chunks: number; mode: 'single' | 'multiline' }
+  whoisChannels(): Promise<string[] | false>
+
+  getHistory(key: string, limit?: number): IrcMessage[]
+  getUsers(channel: string): string[]
```
AlexSc, line 52: Since these aren't returning promises I presume they're coming from local cache -- worth documenting.

AlexSc (changes requested):
Very high level, I think this is worth a forward stress test to make sure we have some notion of how we're going to implement the MCP on top of it, without cluttering it up too much with MCP level concerns.

src/irc-client.ts
```diff
@@ -0,0 +1,64 @@
...
+  // Synchronous socket write — no protocol-level delivery ack for PRIVMSG.
+  say(target: string, text: string): { chunks: number; mode: 'single' | 'multiline' }
+  whoisChannels(): Promise<string[] | false>
+
+  // Served from local cache — no network round-trip. Note: on a freshly-joined channel
+  // these lag the join ack; NAMES (getUsers) and chathistory (getHistory) arrive via
+  // events after join() resolves.
+  getHistory(key: string, limit?: number): IrcMessage[]
+  getUsers(channel: string): string[]
+  // R2 increments on every non-historical inbound message; MCP reads this to build
```
AlexSc: Comments should be timeless.

AlexSc (approved):

PR #145, opened 2026-05-07
fix: stop permbot leaking when parent claude session exits uncleanly

### LINKED ISSUE #138: fix: permbot daemons leak when parent claude session exits
Stale `roost-permbot` processes survive their parent `claude` session and squat the `permbot-<worker>` nick. Next spawn of the same worker name fails registration with IRC 433 (Nickname is already in use) — permbot dies, perm-irc DMs are silently dropped, and the worker blocks indefinitely on permission prompts.

Hit during alpha dogfooding: spawning `watcher` collided with permbots from prior sessions (one from yesterday, one from Monday). The new permbot crashed at startup and the agent silently couldn't request permissions over IRC.

Repro: spawn any worker with `--perm-irc`, kill the parent `claude` without `roost shutdown`, spawn the same worker name again. Second spawn's permbot exits with 433.

Likely cause: `bin/roost` doesn't track parent PID and the permbot has no liveness check on its parent.

**Fix sketch:**
- `bin/roost-permbot` exits when its parent process dies (pidfd / kqueue / poll PPID for the Mac case)
- OR `bin/roost spawn` checks for a stale permbot socket on the same nick and reaps before starting a new one

Either is fine; first is more robust, second is simpler.

**Acceptance:**
- Killing the parent claude session without `roost shutdown` does not leave a permbot running
- Spawning the same-named worker after an unclean exit succeeds without nick collision

[lead-pm]

TeakBuilds:
Closes #138

Two complementary fixes:

**1. Parent-poll (load-bearing)** — `bin/roost-permbot` reads `ROOST_PERM_PARENT_PID` (tmux pane PID passed at spawn) and calls `os.kill(pid, 0)` each select loop iteration. Exits cleanly when the parent is gone.

**2. Stale-reap at spawn (belt-and-suspenders)** — `bin/roost` writes `/tmp/roost-permbot-<nick>.pid` so a fresh spawn can find and kill any stale permbot before registering the same IRC nick. Verifies identity via `ps -o args=` before killing (guards against PID reuse). `stop_permbot` cleans the file on normal shutdown.

`start_permbot` now runs after `tmux new-session` so the pane PID is available as the watchdog target.

[worker-138]

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -118,7 +118,29 @@ start_permbot() {
   local nick="$1"
   local target="$2"
   local data_dir="$3"
+  local parent_pid="${4:-}"
   perm_paths "${data_dir}"
+
+  # Reap any stale permbot from a prior unclean exit before registering the same nick.
+  local well_known_pid="/tmp/roost-permbot-${nick}.pid"
+  if [ -f "${well_known_pid}" ]; then
+    local stale
+    stale="$(cat "${well_known_pid}" 2>/dev/null || true)"
+    if [ -n "${stale}" ] && kill -0 "${stale}" 2>/dev/null; then
+      if ps -o args= -p "${stale}" 2>/dev/null | grep -q "roost-permbot"; then
+        kill "${stale}" 2>/dev/null || true
+        local _j
+        for _j in 1 2 3 4; do
+          kill -0 "${stale}" 2>/dev/null || break
+          sleep 0.25
+        done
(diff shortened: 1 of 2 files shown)


AlexSc:
You know, instead of all of this, would it be possible to have perm-irc sidecar or be a child of the MCP process? That'd tie its lifecycle to the MCP process which does cleanly shutdown.

PR #146, opened 2026-05-07
refactor: migrate state and event handlers into RoostIrcClient (R2/R3)

### LINKED ISSUE #119: refactor: migrate state and event handlers into RoostIrcClient (R2/R3)
Second slice of the RoostIrcClient extraction. Move state variables, their private helpers, and the IRC event handlers from the `createMcpServer` closure into the `RoostIrcClient` class body — together, as a unit.

**State that moves:** `irc_ready`, `hasRegistered`, `join_resolvers`, `part_resolvers`, `multilineMaxLines`, `history`, `unread`, `seenFingerprints`, `channelUsers`. Private helpers: `pushHistory`, `msgFingerprint`, `addFingerprint`, `hasFingerprint`, `ensureChannelSet`. The `@ts-expect-error` irc-framework import moves here — suppression lives once, inside the class.

**Handlers that move (all 12):** `registered`, `join`, `userlist`, `part`, `kick`, `quit`, `nick`, `message`, `batch end draft/multiline`, `batch end chathistory`, `socket close`, `socket error`. Handlers access state as private class members — no temporary public mutators needed.

Where handlers previously called `emitChannelEvent`/`emitMembershipEvent`/`emitSystemEvent`, they now emit typed events from the client (`message`, `membership`, `system`). The MCP shim subscribes to those typed events and calls `pushNotification`.

The `event.nick === NICK` self-event early-return at 8 sites consolidates inside each handler.

**Stays in MCP shim:** `receiveSeq`, `formatUnreadLine`, `unreadSuffix`, `pushNotification` — MCP presentation concerns.

**Acceptance:**
- `RoostIrcClient` holds all listed state and all 12 handlers
- `createMcpServer` wires typed event subscriptions; no direct closure variable access for moved state
- `emitChannelEvent`, `emitMembershipEvent`, `emitSystemEvent` helpers are gone; replaced by typed event subscriptions in the shim
- `@ts-expect-error` import appears once, inside the class
- Tests pass, no behavior change

TeakBuilds:
Closes #119

## What moved

- **State:** `irc_ready`, `hasRegistered`, `join_resolvers`, `part_resolvers`, `multilineMaxLines`, `history`, `unread`, `seenFingerprints`, `channelUsers` → `RoostIrcClientImpl` private fields
- **Private helpers:** `pushHistory`, `msgFingerprint`, `addFingerprint`, `hasFingerprint`, `ensureChannelSet` → private methods
- **All 12 IRC handlers:** `registered`, `join`, `userlist`, `part`, `kick`, `quit`, `nick`, `message`, `batch end draft/multiline`, `batch end chathistory`, `socket close`, `socket error` → `registerHandlers()` in impl
- **`@ts-expect-error` import:** now appears once, inside `irc-client-impl.ts`

## What stays in the shim

`receiveSeq`, `formatUnreadLine`, `unreadSuffix`, `pushNotification`, `emitUnreadSummary` — all MCP presentation concerns.

## Interface additions

`isJoined(channel)` added to `RoostIrcClient` in `irc-client.ts` — surfaced as a real gap when the shim needed to check the already-joined cache cleanly. `channelUsers.has(channel)` → `client.isJoined(channel)`.

`ClientConfig` added to `irc-client.ts` to break a potential circular import (`irc-client-impl.ts` → `irc-server.ts` → `irc-client-impl.ts`). `McpServerConfig` is now a re-exported type alias.

## R4 note

`say()` and `whoisChannels()` are fully implemented here even though they are technically R4 scope — stubs would have been more churn than just doing them. R4 can close as no-op or be downscoped.

## Testing

87 tests pass (all previously skipped integration tests now run with ergo installed). One bug caught and fixed during implementation: `addFingerprint` was missing for historical messages in the `batch end chathistory` handler — would have caused chathistory dedup to fail on part+rejoin of channels with historical-only messages.

### DIFF
--- a/src/irc-client-impl.ts
+++ b/src/irc-client-impl.ts
@@ -1,0 +1,470 @@
+// @ts-expect-error — irc-framework lacks first-class type defs
+import IRC from 'irc-framework'
+import { MULTILINE_LINE_BYTES } from './constants.js'
+import {
+  splitLineForMultiline,
+  newBatchId,
+  reassembleMultilineBatch,
+} from './irc-lib.js'
+import type {
+  RoostIrcClient,
+  ClientConfig,
+  ConnectOpts,
+  IrcMessage,
+  MessageMeta,
+  MembershipExtras,
+  UnreadInfo,
+} from './irc-client.js'
+
+const CAP_CHATHISTORY = 'chathistory'
+
+export class RoostIrcClientImpl implements RoostIrcClient {
+  private readonly nick: string
+  private readonly historySize: number
+  private readonly joinHistoryLines: number
+  private readonly joinHistoryMinutes: number
+  private readonly autoJoin: string[]
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private readonly irc: any
+
+  private irc_ready = false
+  private hasRegistered = false
+  private readonly joinResolvers = new Map<string, Array<(ok: boolean) => void>>()
+  private readonly partResolvers = new Map<string, Array<(ok: boolean) => void>>()
+  private multilineMaxLines = 100
+  private readonly history = new Map<string, IrcMessage[]>()
+  private readonly unread = new Map<string, UnreadInfo>()
+  private readonly seenFingerprints = new Map<string, Set<string>>()
+  private readonly channelUsers = new Map<string, Set<string>>()
+
+  private readonly messageHandlers: Array<(msg: IrcMessage, meta: MessageMeta) => void> = []
+  private readonly membershipHandlers: Array<(kind: 'join' | 'leave' | 'nick', nick: string, channel: string, extras: MembershipExtras) => void> = []
+  private readonly systemHandlers: Array<(kind: 'disconnected' | 'reconnected', content: string) => void> = []
+
+  constructor(config: ClientConfig) {
+    this.nick = config.nick
+    this.historySize = config.historySize
+    this.joinHistoryLines = config.joinHistoryLines
+    this.joinHistoryMinutes = config.joinHistoryMinutes
+    this.autoJoin = config.autoJoin
+    this.irc = new IRC.Client()
+    this.registerHandlers()
+  }
+
+  // ---- Public interface --------------------------------------------------
+
+  connect(opts: ConnectOpts): void {
+    this.irc.requestCap(['draft/multiline', 'labeled-response', CAP_CHATHISTORY, 'server-time'])
+    this.irc.connect({
+      host: opts.host,
+      port: opts.port,
+      nick: opts.nick,
+      username: opts.username ?? opts.nick,
+      gecos: opts.gecos ?? opts.nick,
+      auto_reconnect: opts.autoReconnect ?? false,
+      auto_reconnect_max_retries: opts.autoReconnectMaxRetries,
+    })
+  }
+
+  isReady(): boolean { return this.irc_ready }
+  isJoined(channel: string): boolean { return this.channelUsers.has(channel) }
+
+  async join(channel: string): Promise<boolean> {
+    return new Promise<boolean>((resolve) => {
+      const list = this.joinResolvers.get(channel) ?? []
+      list.push(resolve)
+      this.joinResolvers.set(channel, list)
+      this.irc.join(channel)
+      setTimeout(() => resolve(false), 5000).unref?.()
+    })
+  }
+
+  async leave(channel: string): Promise<boolean> {
+    return new Promise<boolean>((resolve) => {
+      const list = this.partResolvers.get(channel) ?? []
+      list.push(resolve)
+      this.partResolvers.set(channel, list)
+      this.irc.part(channel)
+      setTimeout(() => resolve(false), 5000).unref?.()
+    })
+  }
+
+  say(target: string, text: string): { chunks: number; mode: 'single' | 'multiline' } {
+    if (text.length <= MULTILINE_LINE_BYTES && !text.includes('\n')) {
+      this.irc.say(target, text)
+      return { chunks: 1, mode: 'single' }
+    }
+
+    const id = newBatchId()
+    const logicalLines = text.split('\n')
+    const wireLines: Array<{ body: string; concat: boolean }> = []
+    for (const line of logicalLines) {
+      const chunks = splitLineForMultiline(line)
+      chunks.forEach((chunk, idx) => {
+        wireLines.push({ body: chunk, concat: idx > 0 })
+      })
+    }
+
+    if (wireLines.length > this.multilineMaxLines) {
+      process.stderr.write(
+        `roost-irc[${this.nick}]: multiline target=${target} would emit ${wireLines.length} lines, exceeds server max ${this.multilineMaxLines}; sending anyway\n`,
+      )
+    }
+
+    this.irc.raw('BATCH', `+${id}`, 'draft/multiline', target)
+    for (const { body, concat } of wireLines) {
+      const tagStr = concat ? `batch=${id};draft/multiline-concat` : `batch=${id}`
+      this.irc.connection.write(`@${tagStr} PRIVMSG ${target} :${body}`)
+    }
+    this.irc.raw('BATCH', `-${id}`)
+    process.stderr.write(
+      `roost-irc[${this.nick}]: multiline outbound to ${target} as batch ${id} (${wireLines.length} lines, ${text.length} bytes)\n`,
+    )
+    return { chunks: wireLines.length, mode: 'multiline' }
+  }
+
+  async whoisChannels(): Promise<string[] | false> {
+    return new Promise<string[] | false>((resolve) => {
+      this.irc.whois(this.nick, (event: { channels?: string }) => {
+        if (!event.channels) { resolve([]); return }
+        const list = event.channels
+          .split(' ')
+          .map((ch: string) => ch.replace(/^[@+%&~]+/, ''))
+          .filter(Boolean)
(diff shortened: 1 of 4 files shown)


TeakBuilds (commented):
[reviewer-146] Behavior-preservation pass + simplify on this branch.

Overall: **clean for behavior preservation.** All 12 handlers transcribe correctly. Self-event early-returns at all 8 sites are preserved (join/part/kick/quit/nick/message/multiline-batch sender/chathistory per-msg) and `channelUsers` mutations on self-PART/KICK/QUIT still happen *before* the early return — a subtle property of the original code that was easy to break. Worker-119's chathistory `addFingerprint` fix is positioned correctly: in the new impl `addFingerprint` is the explicit step right before `pushHistory`/`emitMessage` rather than being a side effect of the old `emitChannelEvent`. Every emit path through `emitMessage` now calls `addFingerprint` (live message L378, multiline batch L411, chathistory L453) — no other emit paths exist that could bypass dedupe.

Unread is incremented exactly twice (live msg L380, multiline batch L413) and never on the chathistory path — matches main's `if (!extras.historical)` gate. `MessageMeta` payloads `{}`/`{buffered, chunkCount}`/`{historical}` and `MembershipExtras` `{reason}`/`{newNick}` payload shapes match. Shim's three `client.on(...)` subscribers are the only callers of `pushNotification` for inbound traffic — `grep` confirms zero leftover `emitChannelEvent`/`emitMembershipEvent`/`emitSystemEvent` references in the shim.

## Nit

- `RoostIrcClient.join(channel, force?)` declares a `force` param the impl ignores and the shim consumes *above* the `client.join()` call (`isJoined` check at irc-server.ts:275). Aspirational — either drop the param from the interface or honor it inside the impl in R6. Today it's a footgun for any future caller that goes through the typed seam.
- `RoostIrcClientImpl.quit()` is public but not on the `RoostIrcClient` interface. `test/helpers/mcp-inprocess.ts:56` calls it via the concrete type, so it works. R6 cleanup: add to interface (it's a real lifecycle method) or move test teardown to use a typed shutdown method.
- Slight ordering inconsistency: live-msg and multiline-batch handlers do `pushHistory → addFingerprint`; chathistory does `addFingerprint → pushHistory`. Functionally equivalent (no observer between the two calls) but pick one for consistency.

## Question

- `connect()` now hardcodes the cap list (`draft/multiline`, `labeled-response`, `chathistory`, `server-time`) inside the impl. Main had it on the entrypoint side. Intentional? If R4/R5 ever needs configurable caps (e.g. for a non-ergo server) it'd have to be plumbed back through `ConnectOpts`. Fine for now, just flagging.

LGTM for R2/R3 scope. Ready for review-→-merge once the nits get triaged into R6.

AlexSc:
lead-pm, before I get too deep into this can we get a fresh opus review concentrating on potential risks or issues using this as the substrate for a redone permbot and orchestrator/dispatcher?

TeakBuilds (commented):
[substrate-review-146]

Forward-looking substrate review per @alex's brief: if we re-implemented `bin/roost-permbot` and `bin/orchestrator_poll` on top of `RoostIrcClient`, what would not fit? Read the two Python scripts and `src/irc-client{,-impl}.ts` on this branch. Not a behavior-preservation review (that's done).

## Fits cleanly

- **Single connection per instance.** Both consumers want exactly one IRC socket; impl assumes the same. Match.
- **`join` / `leave` / `isJoined`.** Orchestrator dynamically reconciles its channel set every tick (`bin/orchestrator_poll:1103-1109`). The interface supports this exactly — `isJoined` was even added in this PR for precisely that shape.
- **Subscriber API for inbound.** Permbot needs DMs from one nick; it can subscribe to `'message'` and filter on `msg.isDirect && msg.sender === TARGET`. Orchestrator doesn't read inbound at all and can simply not subscribe. Three event types is a clean subset.
- **Reconnect-rejoin.** The impl's snapshot-rejoin in `registered` (`src/irc-client-impl.ts:249-260`) is what orchestrator hand-rolls today. Free win for the orchestrator.

## Friction

- **`say()` picks the wire shape for you.** Anything `> MULTILINE_LINE_BYTES` or containing `\n` becomes a draft/multiline batch (`src/irc-client-impl.ts:93-125`). Orchestrator deliberately separates `privmsg` (split into single-line chunks) from `comment_privmsg` (multiline batch with explicit fallback to a single fallback line) — see `bin/orchestrator_poll:831-859`. Today the script controls the per-event choice; under this interface it loses that lever. Permbot is fine because every line is short.
- **CAP set hardcoded.** `connect()` requests `draft/multiline + labeled-response + chathistory + server-time` (`src/irc-client-impl.ts:58`). Permbot wants none of those. On the current ergo deployment the server ACKs them and nothing breaks, but a permbot bound to a different server (or one of these caps misbehaving) would drag them along involuntarily.
- **No language story.** Impl is TypeScript; both consumers are Python. Either port both to TS (permbot's unix-socket queue + per-request timeouts is the bulk of its code, not the IRC bits — re-implementing in Node is real work, not a copy/paste) or stand up a Python sibling client. #146 picks neither. That's fine for what #146 *is*, but it's the elephant in the substrate room.
- **No manual `reconnect()`.** Reconnect lives entirely behind `irc-framework`'s `auto_reconnect`. The interface gives `connect()` and not much else. Orchestrator's "lost connection → tear down → reconnect → resume" pattern (`bin/orchestrator_poll:1110-1117`) has to either trust auto-reconnect or re-instantiate the whole client. Workable, but a step back from the explicit `irc.reconnect()` it has today.

## Gaps

1. **`process.exit(1)` if `draft/multiline` isn't ACKed.** `src/irc-client-impl.ts:237-242`. This is the one I'd call a hard blocker for substrate use. A permbot — which sends only `PRIVMSG nick :y/n hint` lines and has zero need for multiline — would inherit a process-suicide on a server that lacks the cap. Substrate code can't unilaterally `process.exit`. Shape: emit a `'system'` event (`'cap-missing'` or similar) and let the consumer decide whether that's fatal.

2. **No registration-failure signal.** Permbot's whole nick-collision path lives at `bin/roost-permbot:257-260` — `432/433/436` is a fatal exit. The interface emits `'system'` for `disconnected` / `reconnected` only. There's no way to learn that registration *failed* vs. is just slow. `isReady()` is a poll, not an event, and never flips true on a 433 — the consumer would just hang on whatever 5s timeout it slapped on top. Shape: `'system'` should grow `'registered'` and `'registration-failed'` (with the numeric reply code), or `connect()` should return a Promise that resolves/rejects on first registration outcome.

3. **`connect()` is fire-and-forget with no completion handle.** Comment on `src/irc-client.ts:44-45` says "use `isReady()` + `on('system')`". For an MCP that wants to start serving before IRC is up, fine. For a permbot daemon that needs to know "are we registered or did the server reject our nick" before it accepts the first hook request on the unix socket, polling `isReady()` is the wrong shape. Same fix as (2).

4. **`say()` has no per-call mode override.** No way to say "force single-line splits, never batch". Orchestrator's `comment_privmsg` already encodes a thoughtful policy ("emit multiline only if it fits server limits, otherwise fallback") that this interface just hides. Shape: optional second arg `{ mode: 'single' | 'multiline' | 'auto' }`, default `auto`.

5. **`quit()` is on the impl class but not the interface.** `src/irc-client-impl.ts:156`. Permbot sends a goodbye `QUIT :daemon shutdown` (`bin/roost-permbot:296`); to do the same through this interface it'd have to downcast or add the method to the interface.

## Recommendation

**Land #146 as-is for the R2/R3 refactor it's actually doing** — the MCP cutover is good, tests are green, behavior is preserved. The substrate ambition is a separate question and shouldn't gate this merge.

**Before either Python script gets ported onto this client, three small shape changes are worth baking in:**
- Drop the `process.exit(1)` on missing multiline; surface as a system event.
- Add registration-success / registration-failure to `'system'` (or return a Promise from `connect()`).
- Either expose `quit()` on the interface or accept that consumers re-instantiate to disconnect.

Anything else (CAP configurability, `say()` mode override, manual `reconnect()`) is nice-to-have and can wait until a real consumer asks for it. The three above are concretely needed by permbot the moment you try to write it on top of this.

Bigger question worth raising explicitly: is the plan (a) port both scripts to TS, (b) build a parallel Python client, or (c) keep the Python scripts as-is and let the TS client serve only the MCP? #146 is a fine refactor under any of those, but the answer changes how much investment the interface deserves.

AlexSc:
My thinking is (a) -- we shouldn't be maintaining 2 (or 3?!) IRC clients

AlexSc (changes requested):
- I would prefer to see the inline types on the event handlers extracted to interfaces.
- Let's consolidate all the stderr writes into a log() helper
- Do we ever pushHistory without addFingerprint? Could consolidate
- Similarly, do we ever pushHistory without updating unread?
- Instead of inlining logic in event handlers I'd like to see it factored out to helper methods

src/irc-client-impl.ts
```diff
@@ -0,0 +1,485 @@
...
+  private readonly nick: string
+  private readonly historySize: number
+  private readonly joinHistoryLines: number
+  private readonly joinHistoryMinutes: number
+  private readonly autoJoin: string[]
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private readonly irc: any
+
+  private irc_ready = false
```
AlexSc: Any reason this isn't camelCase?

AlexSc (changes requested):
lead-pm, can we get a tracking issue to add eslint?

AlexSc (approved):

PR #152, opened 2026-05-07
refactor: rewrite CallTool dispatch as thin wrappers (R5)

### LINKED ISSUE #122: refactor: rewrite CallTool dispatch as thin wrappers (R5)
After R4, the CallTool switch still calls IrcClient methods but the case bodies are still 10-15 lines of inline logic. R5 slims them.

Most cases shrink to 1-3 lines — `channel_message`, `direct_message`, `channel_join`, `channel_leave`, `channel_who`, `channel_ack` each become a call + return. `channel_history` and `channel_list` stay medium (~5-8 lines) because they format output: history slices and formats per-message, list maps `whoisChannels()` result against unread state per channel. That's presentation work that belongs in the shim.

Folds in:
- **D1**: `channel_message` and `direct_message` collapse into a shared helper — both call `client.say(target, text)`, differ only in the target arg. The preview/note/suffix formatting stays as shared logic, not duplicated.
- **L1**: `TOOL_SCHEMAS` lifted to module-level const; ListTools handler becomes `async () => ({ tools: TOOL_SCHEMAS })`.

L13 coordination: if @alex's unread decision lands before this, the suffix lines in channel_message/direct_message may already be gone. Work against the current shape; if L13 has landed, skip the suffix lines.

**Acceptance:**
- `channel_message` and `direct_message` share a helper; duplication from D1 is gone
- `TOOL_SCHEMAS` is a module-level const; ListTools handler is a one-liner
- `channel_join`, `channel_leave`, `channel_who`, `channel_ack` are 1-3 line wrappers
- `channel_history` and `channel_list` are the only cases with non-trivial body; both are under 10 lines
- Tests pass, no behavior change

TeakBuilds:
closes #122

TOOL_SCHEMAS module-level const, ListTools one-liner, handleSay() shared helper for the two say cases, all other cases 3-line wrappers, history/list under 10 lines. no behavior changes.

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -243,33 +243,33 @@ export function createMcpServer(client: RoostIrcClient, config: ClientConfig): {
 
     switch (name) {
       case 'channel_message': {
-        const ch = String(args.channel ?? '')
-        const text = String(args.text ?? '')
+        const ch = args.channel as string
+        const text = args.text as string
         return handleSay(ch, text, `sent to ${ch}`)
       }
       case 'direct_message': {
-        const nick = String(args.nick ?? '')
-        const text = String(args.text ?? '')
+        const nick = args.nick as string
+        const text = args.text as string
         return handleSay(nick, text, `DM to ${nick}`)
       }
       case 'channel_join': {
-        const ch = String(args.channel ?? '').toLowerCase()
+        const ch = (args.channel as string).toLowerCase()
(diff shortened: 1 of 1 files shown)


AlexSc (changes requested):
Since the MCP schema lists arguments as required strings do we really need the extra coercion to non-nullish strings?

AlexSc (approved):

PR #155, opened 2026-05-07
refactor: R6 cleanup — canonicalization, cap-missing, registration outcome

### LINKED ISSUE #123: refactor: cleanup pass — canonicalization, dead defensive code, test helpers (R6)
Final cleanup slice after R3 (handlers in IrcClient) and R5 (shim thinned).

**L5 — drop defensive `set.add(NICK)` in userlist handler.** Server sends `RPL_NAMREPLY/ENDOFNAMES` only after a successful JOIN, which guarantees our nick is in the list. The defensive add at irc-server.ts:631 protects against a server contradiction. Delete it and the stale comment.

**L14 — commit to a canonicalization rule.** Two rules, applied consistently:
- Every public `RoostIrcClient` method that accepts a channel arg (e.g. `join`, `leave`, `getHistory`, `getUsers`, `ackUnread`) lowercases on entry.
- Every handler that reads a channel name from an IRC event (`event.channel`, `event.target`) lowercases before writing into any state Map (`history`, `unread`, `channelUsers`, `seenFingerprints`).

Assumption: lowercase is correct for ergo. RFC 1459 case-mapping (`{|}^` ↔ `[\\]~`) is out of scope; revisit if Q6 lands a non-ergo backend.

**L10 — test helpers.** `startMcp` (subprocess) and `startMcpInProcess` both construct `RoostIrcClient` directly after the extraction. Their structural difference (process boundary vs in-process) is real and should be visible; any setup that's now identical can be extracted to a shared helper or left parallel with a comment naming the difference.

**From R2/R3 review (#146 reviewer-146):**
- **`force` param on `join()` is aspirational.** Declared on the interface, ignored by the impl, consumed only in the shim's `isJoined` precheck. Either drop from the interface or honor inside the impl.
- **`quit()` is on the impl but not the interface.** `test/helpers/mcp-inprocess.ts:56` calls it via the concrete type. Add to the interface (real lifecycle method) or move test teardown to a typed shutdown.
- **`pushHistory`/`addFingerprint` ordering inconsistency.** Live-msg and multiline-batch handlers do `pushHistory → addFingerprint`; chathistory does `addFingerprint → pushHistory`. Functionally equivalent — pick one.

**From R2/R3 substrate review (#146 substrate-review-146):**
@AlexSc confirmed plan: port `bin/roost-permbot` and `bin/orchestrator_poll` to TS on top of `RoostIrcClient` (single IRC client codebase, no parallel Python). Three interface shape changes are blockers for those ports:
- **Drop `process.exit(1)` on missing `draft/multiline` cap** (`src/irc-client-impl.ts:237-242`). Library code can't unilaterally exit. Surface as a `'system'` event (e.g. `'cap-missing'`) and let the consumer decide.
- **Add registration outcome to `'system'` events** (`'registered'` / `'registration-failed'` with the numeric reply code), or have `connect()` return a Promise that resolves/rejects on first registration outcome. Permbot bails on 432/433/436; today the interface doesn't surface that.
- **`quit()` on the interface** — same nit as above, doubly motivated by substrate use.

**Acceptance:**
- `set.add(NICK)` and its comment deleted from userlist handler
- All state Maps keyed on lowercased channel names; all public methods lowercase channel args on entry
- `grep -n "event\.channel\|event\.target"` in IrcClient shows lowercasing at every write site
- Test helpers construct `RoostIrcClient` directly; shared setup is consolidated or visibly parallel
- `force` either dropped or wired through impl
- `quit()` on the interface or replaced
- handler emit ordering uniform
- `process.exit(1)` on missing cap removed; surfaced as `'system'` event
- registration outcome (success/failure) surfaced on `'system'` or via `connect()` Promise
- Tests pass

TeakBuilds:
Closes #123.

R6 final cleanup pass on top of R1–R5.

**L5** — dropped `set.add(NICK)` from userlist handler. Server guarantees our nick is in NAMES after a successful JOIN.

**L14** — lowercase canonicalization committed: all public methods lowercase channel/key args on entry; all handlers canonicalize `event.channel`/`event.target` before any state-Map write. DM keys use `sender.toLowerCase()`.

**quit()** — added to `RoostIrcClient` interface.

**cap-missing** — `process.exit(1)` removed from `parseMultilineCap`; replaced with `emitSystem('cap-missing', ...)`. `parseMultilineCap` returns `bool`; `handleRegistered` bails on false.

**Registration outcome** — option A (system events). `emitSystem('registered', { code: 1, nick })` on first successful registration. 432/433/436 numeric listeners emit `emitSystem('registration-failed', { code })` while `!ircReady`. `SystemKind`/`SystemContent` exported from `irc-client.ts`; `irc-server.ts` stringifies structured content for `pushNotification`.

**L10** — one-liner comments on `startMcp`/`startMcpInProcess` naming the structural difference (process-boundary vs explicit-quit teardown).

[worker-123]

### DIFF
--- a/src/irc-client-impl.ts
+++ b/src/irc-client-impl.ts
@@ -285,16 +285,16 @@ export class RoostIrcClientImpl implements RoostIrcClient {
   // ---- IRC event handlers ------------------------------------------------
 
   private handleRegistered(): void {
-    this.ircReady = true
     this.log('registered with the IRC server')
     if (!this.parseMultilineCap()) return
+    this.ircReady = true
     this.logChathistoryCap()
     if (this.hasRegistered) {
       this.handleReconnect()
       return
     }
     this.hasRegistered = true
-    this.emitSystem('registered', { code: 1, nick: this.nick })
+    this.emitSystem('registered', { nick: this.nick })
     for (const ch of this.autoJoin) {
       this.irc.join(ch)
       this.log(`auto-joining ${ch}`)
--- a/src/irc-client-impl.ts
+++ b/src/irc-client-impl.ts
@@ -14,6 +14,8 @@ import type {
   MessageMeta,
   MembershipExtras,
   UnreadInfo,
+  SystemKind,
+  SystemContent,
 } from './irc-client.js'
 
 const CAP_CHATHISTORY = 'chathistory'
@@ -66,7 +68,7 @@ export class RoostIrcClientImpl implements RoostIrcClient {
 
   private readonly messageHandlers: Array<(msg: IrcMessage, meta: MessageMeta) => void> = []
   private readonly membershipHandlers: Array<(kind: 'join' | 'leave' | 'nick', nick: string, channel: string, extras: MembershipExtras) => void> = []
-  private readonly systemHandlers: Array<(kind: 'disconnected' | 'reconnected', content: string) => void> = []
+  private readonly systemHandlers: Array<(kind: SystemKind, content: SystemContent) => void> = []
 
   constructor(config: ClientConfig) {
     this.nick = config.nick
@@ -94,7 +96,7 @@ export class RoostIrcClientImpl implements RoostIrcClient {
   }
 
   isReady(): boolean { return this.ircReady }
-  isJoined(channel: string): boolean { return this.channelUsers.has(channel) }
+  isJoined(channel: string): boolean { return this.channelUsers.has(channel.toLowerCase()) }
 
   async join(channel: string): Promise<boolean> {
     channel = channel.toLowerCase()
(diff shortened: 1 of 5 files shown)


TeakBuilds (commented):
[reviewer-155]

Checklist all passes. Notes below.

**L14 canonicalization** — clean. Every handler opens with `const channel = event.channel.toLowerCase()` (join/userlist/part/kick) or computes `isDirect ? event.nick.toLowerCase() : event.target.toLowerCase()` (message/multiline). Public method entrances (isJoined, getHistory, getUsers, ackUnread) all lowercase on entry. `recordMessage` uses `msg.channel` which is already lowercased at construction. No missed sites.

**L5** — `set.add(this.nick)` and its comment fully removed from userlist handler.

**cap-missing** — `process.exit(1)` gone. `parseMultilineCap` returns bool; `handleRegistered` bails on false after emitting `cap-missing`. Good.

**registration outcome** — `registered`/`registration-failed` events fire correctly. 432/433/436 wired with `!ircReady` guard. Numerics passed through on both events.

**quit()** — on the interface.

**L10 test helpers** — one-liner contrast comments, not merged.

**DM nick keys** — all three sites (handleMessage, handleMultilineBatch, parseChathistoryBatch) use `sender.toLowerCase()` for DMs, not `event.target`.

---

**Minor items (non-blocking):**

1. `ircReady` is set `true` before `parseMultilineCap()` is checked — so `isReady()` returns true even when `cap-missing` fires. The consumer gets the signal but could be confused by `isReady() === true` alongside it. Probably acceptable given the old behavior was exit(1).

2. `SystemContent` is `string | { code: number; nick?: string }` with a comment mapping kinds to shapes, but TypeScript won't enforce it. Nothing stops passing a string for `registered`. A discriminated union keyed by SystemKind would catch this. Low stakes for internal code.

3. `{ code: 1 }` on the `registered` event is hardcoded, not read from the IRC 001 wire. If the intent is just "we registered successfully", `nick` alone is sufficient and `code` is noise. If it's meant as the IRC numeric, it should come from the actual event.

4. DM channel-key pattern (`isDirect ? sender.toLowerCase() : rawTarget.toLowerCase()`) appears 3x. A tiny helper would DRY it up, but three simple one-liners is fine by house rules.

AlexSc (approved):

PR #159, opened 2026-05-07
fix: pre-empt join/part resolvers on socket close

### LINKED ISSUE #125: fix: socket close leaves join/part resolvers on 5s timer (C1)
When the socket drops mid-`channel_join`, the resolver stays alive until its 5s timer fires. The caller gets `'join #foo timed out'` instead of a disconnection signal. Same shape for `channel_leave`.

The `socket close` handler at irc-server.ts:813-817 only clears `irc_ready` and emits a system event — it doesn't pre-empt the pending resolvers.

Fix: in the `socket close` handler, walk both `join_resolvers` and `part_resolvers`, resolve each entry with `false` immediately, then clear the maps. Resolver callsites already handle `false` — no other changes needed.

**Acceptance:**
- `socket close` pre-empts all pending join/part resolvers with `false`
- A `channel_join` in progress when the socket drops returns an error result immediately, not after 5s
- Tests pass

TeakBuilds:
closes #125

in `handleSocketClose()`, walk both resolver maps, resolve each with `false`, then clear. callers already handle `false`. stale `setTimeout`s still fire but calling `resolve()` on a settled promise is a no-op per spec (noted in comment).

3 new tests in `disconnect-cache.test.ts` covering join/part resolver pre-emption and map clearing.

### DIFF
--- a/src/irc-client-impl.ts
+++ b/src/irc-client-impl.ts
@@ -495,6 +495,11 @@ export class RoostIrcClientImpl implements RoostIrcClient {
       this.pendingRejoinChannels = [...this.channelUsers.keys()].sort()
       this.channelUsers.clear()
     }
+    // Pre-empt pending resolvers; stale setTimeouts still fire but calling resolve() again is a no-op.
+    for (const list of this.joinResolvers.values()) for (const r of list) r(false)
+    this.joinResolvers.clear()
+    for (const list of this.partResolvers.values()) for (const r of list) r(false)
+    this.partResolvers.clear()
     this.ircReady = false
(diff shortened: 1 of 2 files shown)


TeakBuilds (commented):
[reviewer-159] LGTM. ran /simplify (reuse, quality, efficiency) — all clean.

- comment on stale setTimeouts earns its place (non-obvious Promise spec invariant)
- two near-duplicate resolver loops are too small to abstract
- stale timers are already .unref()'d and bounded to 5s; no memory concern
- test access via existing `as any` pattern, no new lint risk
- CI green, 36 tests pass

AlexSc (approved):

Issue #153, opened 2026-05-07
port bin/roost-permbot to TS on top of RoostIrcClient


TeakBuilds:
After R6 lands the substrate fixes (drop process.exit on missing multiline cap → 'system' event; surface registration outcome; quit() on interface), `bin/roost-permbot` becomes portable to TS.

Goal: single IRC client codebase. No parallel Python.

**Scope:**
- Port permbot's IRC role (DM y/n permission prompts to perm-target, listen for response on per-spawn unix socket) to TypeScript on top of `RoostIrcClient`.
- Preserve current external interface: nick `permbot-<spawn-nick>`, unix socket at `${ROOST_DATA_DIR}/permbot.sock`, lifecycle tied to MCP child process (#148).
- Delete `bin/roost-permbot` (Python) once TS port is wired in `bin/roost`.

**Blocked by:** #123 (R6 substrate fixes)

**Out of scope:** orchestrator port (#TBD-orchestrator), behavioral changes to permbot prompts.

Issue #148, opened 2026-05-07
permbot: tie lifecycle to MCP child process (replaces parent-PID poll)


TeakBuilds:
Followup to #138 / PR #145.

#145 fixes the leak via parent-PID polling + well-known-PID stale reap. Works, ~50 LOC of bash + python.

[@AlexSc](https://github.com/AlexSc) suggested a structurally cleaner alternative: have permbot be a child of the MCP server process (or spawned by it) instead of a sibling daemon. The MCP shuts down cleanly when claude exits, so permbot inherits that lifecycle for free — no polling, no stale-PID reap.

**Why deferred:** requires changes to \`src/irc-server.ts\`, which is frozen behind the active R2-R6 refactor. Once R6 lands, this becomes the right path.

**Sketch:**
- MCP server (or a small wrapper) spawns permbot as a subprocess on startup.
- permbot's stdin/stdout/stderr can be tied to the MCP for clean ownership; or just rely on subprocess-on-parent-exit semantics.
- Drop \`bin/roost-permbot\`'s parent-PID polling loop.
- Drop \`bin/roost\`'s well-known-PID stale-reap block.
- \`bin/roost\` no longer manages permbot lifecycle at all.

**Worker-138 also identified an inner-zsh-trap variant** (do permbot startup + \`trap 'kill \$PERM_PID' EXIT\` inside the tmux'd zsh) that doesn't touch src/irc-server.ts. If we want permbot decoupled from MCP for any reason, that's the alternate path.

After R6, evaluate both and pick.

PR #160, opened 2026-05-07
feat: port bin/roost-permbot to TypeScript (issue #153)

TeakBuilds:
Replaces the 327-line Python permbot with a TS port on top of `RoostIrcClient`.

## What changed
- `src/permbot.ts` — new TS daemon (exports `startPermbot` for testability)
- `bin/roost-permbot` — replaced Python with a 4-line bash shim calling `bun`
- `bin/roost` line 131 — grep pattern updated to `grep -qE "(roost-permbot|src/permbot\.ts)"` (matches both old Python process args and new bun args; tighter than `grep "permbot"` to avoid false-positives on editor processes)
- `test/permbot.test.ts` — 4 in-process unit tests: enqueue→dispatch, timeout, serial queue, shutdown drain

## External interface: unchanged
nick `permbot-<spawn-nick>`, socket `${ROOST_DATA_DIR}/permbot.sock`, same JSON wire protocol `{summary, timeout}` → `{reply}|{timeout:true}|{error}`, same env vars.

## Notes
- `disconnected` → fatal: matches Python. transient ergo restart kills permbot until session ends. acceptable for alpha; revisit with #148.
- `cap-missing` subscribed and logged/ignored explicitly.
- parent-PID polling preserved: `setInterval` + `process.kill(ppid, 0)` / `ESRCH` check.

[worker-153]


TeakBuilds (commented):
[reviewer-160]

Side-by-side vs `main:bin/roost-permbot`. Net: contract preserved, ship-ready aside from one small behavioral nit worth a quick decision.

## Behavioral drift

**Indented summary lines get left-stripped.** `src/permbot.ts:60`:
```ts
...entry.summary.split('\n').map(l => l.trim()).filter(Boolean)
```
Python used `ln.rstrip()` (right-strip only). If a hook ever sends a summary like `"Bash command:\n  rm -rf /tmp"`, Python preserves the 2-space indent, TS flattens it. Probably not load-bearing today (current hooks don't appear to indent), but it's a silent format change and `.rstrip()` → `.replace(/\s+$/, '')` is a one-character fix. Your call whether to match exactly.

## Minor differences (non-blocking, flagging for awareness)

- **Sender match is now case-insensitive** (`msg.sender.toLowerCase() === target.toLowerCase()`). Python compared exact. TS is arguably more correct per IRC nick semantics, but it's a behavior change.
- **`timeout: 0` no longer immediately fires.** `Number(req.timeout) > 0 ? ... : 30` defaults a zero/negative timeout to 30s; Python passed 0 through and tripped the deadline immediately. Edge case, unlikely to matter.
- **QUIT message differs.** Python sent `QUIT :daemon shutdown`; `client.quit()` uses irc-framework's default. Cosmetic.
- **Required-env failure exits with 2** instead of Python's 1. Cosmetic.

## Contract checks (all match)

- Wire JSON shapes (`{summary,timeout}` → `{reply}|{timeout:true}|{error}`): same.
- DM strings (`[{worker}] permission requested:`, body, `reply y/n`, `late — request already timed out (no in-flight prompt)`): identical.
- Socket perms 0o600: ✓ (`fs.chmodSync` in listen callback).
- Parent-PID poll: 1s `setInterval`, ESRCH→stop, other errors swallowed (matches Python's EPERM-as-alive). ✓
- Shutdown order (drain in-flight → drain queue → quit → close server → unlink): matches.
- System events: `registered`/`registration-failed`/`disconnected` all wired to fatal-or-log; `cap-missing` subscribed and explicitly ignored as PR notes.

## Tests

Cover the four stated cases (dispatch/timeout/queue/drain). Mock client is clean — no `any` casts. No ergo dependency. ✓

## Bash shim & `bin/roost` grep

- `bin/roost-permbot` shim: `BASH_SOURCE[0]`-based `ROOST_DIR` resolution is cwd-independent. `exec` keeps signal delivery clean. ✓
- `bin/roost:130` grep `(roost-permbot|src/permbot\.ts)` matches `bun /…/src/permbot.ts` after the exec. Broader form per your authorization. ✓

## Style nits (ignore freely)

- `dlog(logPath, nick, msg)` is only ever called from a `log` closure that already captured both — could just inline or hoist `log` to module scope.
- `Number(req.timeout)` evaluated twice in one ternary. Trivial.

LGTM modulo the rstrip-vs-trim call.

src/permbot.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  function maybeDispatch(): void {
+    if (inFlight !== null || queue.length === 0 || shuttingDown) return
+    const entry = queue.shift()!
+    const lines = [
+      `[${worker}] permission requested:`,
+      ...entry.summary.split('\n').map(l => l.trimEnd()).filter(l => l.trim()),
+      'reply y/n',
+    ]
+    for (const line of lines) client.say(target, line)
```
AlexSc: I'm going to request behavior changes now that we have a real IRC client.

Let's send this as multiline. Poor lead-pm has been getting blasted with a half dozen messages on every permission request. We can condense that to one now. Saves context and turns.

src/permbot.ts
```diff
@@ -0,0 +1,224 @@
...
+    log(`in-flight to ${target} (${lines.length} lines, timeout ${entry.timeout}s)`)
+    const timer = setTimeout(() => {
+      log('in-flight timed out')
+      inFlight = null
+      respond(entry.socket, { timeout: true })
+      maybeDispatch()
```
AlexSc, line 78: What does calling maybeDispatch() here do? I seem to be missing something

AlexSc (changes requested):
lead-pm: Also I'd like to port the irc-permission-prompt script to TS, since it too contains an IRC client.

src/permbot.ts
AlexSc: Bumping this

src/permbot.ts
AlexSc: Bumping this

src/permbot.ts
```diff
@@ -0,0 +1,224 @@
...
+      buf += chunk.toString('utf8')
+      const nl = buf.indexOf('\n')
+      if (nl === -1) return
+      const line = buf.slice(0, nl)
+      buf = buf.slice(nl + 1)
+      let req: { summary?: unknown; timeout?: unknown }
+      try {
+        req = JSON.parse(line) as { summary?: unknown; timeout?: unknown }
+      } catch (e) {
+        respond(socket, { error: `bad request json: ${e}` })
```
AlexSc, line 133: Log bad requests

src/permission-prompt.ts
```diff
@@ -0,0 +1,254 @@
...
+    client.on('system', (kind) => {
+      // 'registered' fires when multiline cap is available (real ergo);
+      // 'cap-missing' fires otherwise — both mean the connection is ready.
+      if ((kind === 'registered' || kind === 'cap-missing') && !sent) {
+        sent = true
+        client.say(PERM_TARGET, `[${WORKER}] terminal fallback: ${reason}`)
+        for (const ln of summary.split('\n')) {
+          if (ln.trim()) client.say(PERM_TARGET, `  ${ln.trimEnd()}`)
+        }
+        client.say(PERM_TARGET, '(worker blocked on terminal — no one watching)')
```
AlexSc: Add a suggestion to use roost tail and roost send to debug

Let's also trigger this if we go for 10 minutes without a perm resolution

AlexSc (changes requested):

AlexSc (approved):

PR #163, opened 2026-05-07
fix: server-time cap validated on registration (#126)

### LINKED ISSUE #126: fix: server-time cap requested but not validated on registration (C2)
`server-time` is requested at startup but never checked after registration — unlike `draft/multiline`, which exits if the server doesn't grant it (irc-server.ts:570-575).

If `server-time` is silently denied, live messages fall back to `new Date().toISOString()` while chathistory replay reads the server-stamped time via `getServerTime()`. Same message, two different `ts` values → fingerprint mismatch → dedupe fails → duplicate emitted. Exact shape of issue #44.

Fix: mirror the multiline gate — after the `registered` event, check `enabled.includes('server-time')` and call `process.exit(1)` with a stderr message if it's missing.

**Acceptance:**
- Registration handler exits with a clear error if `server-time` cap is not granted
- Behavior matches the existing `draft/multiline` gate pattern
- Tests pass

TeakBuilds:
closes #126

## what

after registration, check `enabled.includes('server-time')`. if absent, emit `system 'cap-missing'` with content `'server-time cap not enabled by server'` and bail — ircReady stays false. consumer decides what to do.

mirrors the existing draft/multiline gate exactly. no process.exit.

## test

unit test in disconnect-cache.test.ts (no ergo needed): inject caps without server-time, call handleRegistered directly, assert cap-missing fires and ircReady stays false.

### DIFF
--- a/src/irc-client-impl.ts
+++ b/src/irc-client-impl.ts
@@ -290,6 +290,12 @@ export class RoostIrcClientImpl implements RoostIrcClient {
   private handleRegistered(): void {
     this.log('registered with the IRC server')
     if (!this.parseMultilineCap()) return
+    const enabled: string[] = this.irc.network?.cap?.enabled ?? []
+    if (!enabled.includes('server-time')) {
+      this.log(`server-time NOT enabled (server caps: ${enabled.join(',') || '(none)'})`)
+      this.emitSystem('cap-missing', 'server-time cap not enabled by server')
+      return
+    }
     this.ircReady = true
     this.logChathistoryCap()
     if (this.hasRegistered) {
--- a/test/disconnect-cache.test.ts
+++ b/test/disconnect-cache.test.ts
@@ -61,6 +61,23 @@ describe('channelUsers cache cleared on disconnect', () => {
   })
 })
 
+describe('cap-missing system event on registration', () => {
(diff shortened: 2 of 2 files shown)


TeakBuilds (commented):
[reviewer-163] simplify + eyeball pass — clean, one note.

CI green. No bugs, no leaks, no efficiency issues.

**One mild quality flag:** the new inline check in `handleRegistered` re-reads `this.irc.network?.cap?.enabled ?? []` independently — same line already exists in `parseMultilineCap` (line 308). The structural asymmetry is the real thing: `draft/multiline` lives in an extracted method that returns bool, `server-time` is inlined. Not a blocker for this PR, but if a third cap gets added someone will copy-paste a fourth time. Worth considering `parseServerTimeCap(): boolean` to match the existing pattern, or a generic `requireCap(name): boolean` helper.

Stringly-typed `'cap-missing'` / `'server-time cap not enabled by server'` is pre-existing debt, not new.

Everything else checks out: inline `enabled.includes('server-time')` (not a misleading parse fn), payload clearly names the missing cap, returns before `ircReady = true`, test exercises it without ergo.

AlexSc (approved):

Issue #147, opened 2026-05-07
perm_paths/kill-loop dedup in bin/roost (followup to #138)


TeakBuilds:
[reviewer-145] flagged on #145, deferred:

- \`cmd_spawn\` calls \`perm_paths "\${ROOST_DATA_DIR}"\` redundantly inside the \`perm_irc\` block — already called unconditionally a few lines above and again in \`start_permbot\`.
- The kill-wait-SIGKILL loop is duplicated between the new stale-reap block and \`stop_permbot\`. A \`_kill_gracefully <pid>\` helper (~8 lines) would unify them.
- \`/tmp/roost-permbot-\${nick}.pid\` path is constructed in two functions — could centralize.

None of these block #138 from merging. Small bash-only cleanup, ~30 LOC.

PR #165, opened 2026-05-07
feat: tie permbot lifecycle to MCP child process (issue #148)

### LINKED ISSUE #148: permbot: tie lifecycle to MCP child process (replaces parent-PID poll)

TeakBuilds:
Closes #148. Supersedes #147 (perm_paths/kill-loop dedup is gone — feel free to close that as superseded once this lands).

## What changed

- `src/irc-server.ts`: when `ROOST_PERM_SOCK` + `ROOST_PERM_TARGET` are set in env, the MCP entrypoint spawns `src/permbot.ts` as a child via `Bun.spawn` (stdin=ignore, stdout=ignore — MCP owns stdout for JSON-RPC — stderr=inherit). Derives `ROOST_PERM_NICK=permbot-${NICK}`, `ROOST_PERM_WORKER=${NICK}`, `ROOST_PERM_DEBUG_LOG=${DATA_DIR}/permbot.log`. SIGTERMs the child on `process.on('exit')`, SIGINT, and SIGTERM.
- `src/permbot.ts`: drops `parentPid` config field and `ROOST_PERM_PARENT_PID` env var. Captures `process.ppid` at startup and polls for change (1s); when MCP exits → kernel reparents → ppid changes → permbot exits. Bounded orphan window of one poll interval. Standalone-launched permbots get the same lifecycle automatically.
- `bin/roost`: deletes `start_permbot`, `stop_permbot`, `perm_paths`, the `/tmp/roost-permbot-${nick}.pid` stale-reaper, and the per-session `permbot.pid` file. Keeps env passing via `tmux -e` so MCP picks up the perm vars at startup. `cmd_shutdown` just `rm -rf`s the data dir; permbot exits on its own when its parent (the MCP, riding the tmux pane) dies.
- `test/permbot.test.ts`: drops the now-vestigial `parentPid: null` from each `startPermbot` call. Existing queue/dispatch/timeout/nudge/shutdown-drain tests remain valid because the `startPermbot` interface is unchanged.

## Lifecycle

```
zsh -l (tmux pane)
  └─ claude
      └─ bun src/irc-server.ts (MCP, nick: ${WORKER})
          └─ bun src/permbot.ts (permbot-${WORKER})
```

When claude exits (cleanly or via SIGKILL), MCP gets stdio EOF and exits. Permbot reparents to launchd/init, the ppid watcher fires within ~1s, and permbot exits. No stale daemon, no nick collision on next spawn.

## Test plan

- [x] `bun test` — 116 pass / 0 fail, no regressions in queue/dispatch/timeout/nudge/shutdown.
- [ ] Manual: `roost spawn --perm-irc` end-to-end; verify permbot is a child of bun-MCP via `pstree`.
- [ ] Manual: kill the parent claude with SIGKILL; verify permbot exits within ~1s without leaving a `/tmp/roost-permbot-*.pid` file.
- [ ] Manual: re-spawn the same nick; no 433 nick-in-use.

## Notes

- Lifecycle is now an integration concern, not unit-tested. The `startPermbot` interface is stable so existing unit tests still cover the daemon logic; the MCP-spawn + ppid-reparent path is exercised by manual roost spawn/kill testing.
- `disconnected → fatal` (the transient-ergo-restart issue PR #160 flagged) is out of scope — separate concern.

[worker-148]

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -350,6 +350,46 @@ if (import.meta.main) {
     }
   }
 
+  // Spawn permbot as our child when the side-daemon is requested. Permbot's
+  // lifecycle then rides this MCP's lifecycle: when claude exits, we exit
+  // (stdio EOF), permbot detects ppid change and exits. No external pidfiles,
+  // no stale-reap on next spawn.
+  const PERM_SOCK = process.env['ROOST_PERM_SOCK'] ?? ''
+  const PERM_TARGET = process.env['ROOST_PERM_TARGET'] ?? ''
+  let permbotProc: import('bun').Subprocess | null = null
+  if (PERM_SOCK && PERM_TARGET) {
+    const permbotPath = new URL('./permbot.ts', import.meta.url).pathname
+    const permbotEnv: Record<string, string> = {
+      ...(process.env as Record<string, string>),
+      ROOST_PERM_NICK: `permbot-${NICK}`,
+      ROOST_PERM_SOCK: PERM_SOCK,
+      ROOST_PERM_TARGET: PERM_TARGET,
+      ROOST_PERM_WORKER: NICK,
+    }
+    if (!permbotEnv['ROOST_PERM_DEBUG_LOG'] && DATA_DIR) {
+      permbotEnv['ROOST_PERM_DEBUG_LOG'] = `${DATA_DIR}/permbot.log`
+    }
+    // stdout=ignore: MCP owns parent stdout for JSON-RPC protocol; permbot
+    // writes there would corrupt it. stderr=inherit lets permbot logs surface
+    // alongside MCP logs in the tmux pane.
+    permbotProc = Bun.spawn([process.execPath, permbotPath], {
+      env: permbotEnv,
+      stdin: 'ignore',
+      stdout: 'ignore',
+      stderr: 'inherit',
+    })
+    process.stderr.write(`roost-irc[${NICK}]: spawned permbot child (pid ${permbotProc.pid})\n`)
+  }
+
+  const killPermbot = () => {
+    if (permbotProc && permbotProc.exitCode === null) {
+      try { permbotProc.kill('SIGTERM') } catch { /* ignore */ }
+    }
+  }
+  process.on('exit', killPermbot)
+  process.on('SIGINT', () => { killPermbot(); process.exit(130) })
+  process.on('SIGTERM', () => { killPermbot(); process.exit(143) })
+
   const clientConfig: ClientConfig = {
     nick: NICK,
     autoJoin: AUTO_JOIN,
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -359,21 +359,16 @@ if (import.meta.main) {
   let permbotProc: import('bun').Subprocess | null = null
   if (PERM_SOCK && PERM_TARGET) {
     const permbotPath = new URL('./permbot.ts', import.meta.url).pathname
-    const permbotEnv: Record<string, string> = {
-      ...(process.env as Record<string, string>),
-      ROOST_PERM_NICK: `permbot-${NICK}`,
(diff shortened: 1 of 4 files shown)


src/irc-server.ts
```diff
@@ -350,6 +350,41 @@ if (import.meta.main) {
...
+  }
+
+  const killPermbot = () => {
+    if (permbotProc && permbotProc.exitCode === null) {
+      try { permbotProc.kill('SIGTERM') } catch { /* ignore */ }
+    }
+  }
+  process.on('exit', killPermbot)
+  process.on('SIGINT', () => { killPermbot(); process.exit(130) })
+  process.on('SIGTERM', () => { killPermbot(); process.exit(143) })
```
AlexSc, line 386: What's with these exit codes? Mostly curious. Also, I've no issue with a graceful shutdown but I want to confirm that the child process should naturally terminate when the MCP does. Bun spawn isn't detaching it, right?

AlexSc (commented):
Just checking a thing, please re-request review on answering.

src/irc-server.ts
TeakBuilds: [worker-148] two parts:

**Exit codes:** convention is `128 + signal_number` for processes that exit due to a signal. 130 = SIGINT (2), 143 = SIGTERM (15). Without an explicit code, `process.exit()` defaults to 0, which would lie to whatever invoked us about *why* we exited (`echo $?` shows 0 instead of the signal-derived code). Cosmetic but standard. Could drop them and let the default be 0 if you'd rather.

**Detaching:** no, `Bun.spawn` does not detach by default (`detached: false`). The permbot child stays in the MCP's process group, so:

- Ctrl-C in the tmux pane → SIGINT to the whole pgroup → both die. Our handler is belt-and-suspenders.
- Clean MCP exit (claude exits → stdio EOF on the JSON-RPC pipe → MCP returns from `await mcp.connect(...)`-driven main): no signal is sent automatically. That's why we wire `process.on('exit', killPermbot)` — synchronous SIGTERM to the child before our event loop tears down.
- MCP SIGKILL: can't be intercepted. Permbot orphans, kernel reparents to launchd/init, `process.ppid` changes, watcher fires within ~1s and calls `stop()`. That's the safety net the issue was originally about.

So the three layers compose: process-group propagation for terminal signals, explicit forward for graceful exit, ppid watcher for unclean death.

AlexSc: Great, thanks for the clarification. No changes needed, good work on the defense in depth.

AlexSc (approved):

PR #169, opened 2026-05-07
feat: append IRC-reply reminder to inbound messages (#136)

### LINKED ISSUE #136: feat: periodic IRC-reply reminder attached to inbound messages
Agents can slip from IRC replies into their TUI mid-session — not from gradual drift but instantly, one message turning into silence. A PostToolUse counter misses this; the reminder needs to land when the failure actually happens: when a message arrives and a reply is warranted.

**Mechanism:** attach the reminder to inbound IRC notifications in `pushNotification`. On the first message received in a session, always append the reminder. On subsequent messages, append with ~1-in-5 to 1-in-10 probability (randomness avoids pattern-match-ignoring). No filtering required — the nudge is about having it in context, not reacting to any specific message. If filtering self-events is trivial, do it; otherwise skip it.

**Reminder text:** generic nudge, not context-aware. "Substantive replies should be posted to IRC." The failure mode is forgetting IRC exists, not failing to identify what to reply to.

**Out of scope:** context-aware mention detection (that's the beta mention-flag issue); PostToolUse cadence tracking.

**Acceptance:**
- First inbound message in a session always carries the reminder
- Subsequent messages carry it with ~1-in-5 to 1-in-10 probability
- Reminder text is a generic "reply on IRC" nudge appended to notification content

TeakBuilds:
## Summary
- Append a generic "Substantive replies should be posted to IRC." nudge to inbound chat notifications in `src/irc-server.ts` `pushNotification` path.
- First inbound message of a session always carries it; subsequent messages get it with probability 1/7 (midpoint of the 1/5–1/10 range from the issue).
- Historical replay is skipped (does not count toward the counter, no reminder appended).
- Self-events were already filtered by `irc-client-impl`; no extra work there.
- `meta.reminder='true'` is set when the reminder is appended (test affordance).

Closes #136.

## Test plan
- [x] `test/irc-server.test.ts` — added 3 tests: first-channel-message, first-DM, and historical-replay. All 3 pass locally (`bun test -t reminder`).
- [ ] CI runs full suite.

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -33,6 +33,11 @@ import type { RoostIrcClient, ClientConfig, UnreadInfo } from './irc-client.js'
 
 const SOURCE_NAME = 'roost-irc'
 
+const REPLY_REMINDER = 'Substantive replies should be posted to IRC.'
+// 1/7 — midpoint of the 1/5–1/10 range from #136. Random rate avoids the
+// pattern-match-and-ignore failure mode of a fixed cadence.
+const REMINDER_PROBABILITY = 1 / 7
+
 // Re-export ClientConfig under the legacy name for callers that import McpServerConfig.
 export type { ClientConfig as McpServerConfig } from './irc-client.js'
 
@@ -136,12 +141,9 @@ export function createMcpServer(client: RoostIrcClient, config: ClientConfig): {
   // millisecond timestamp (the original bug behind reassembly).
   let receiveSeq = 0
 
-  // Inbound chat message counter — the first arriving message always carries
-  // the IRC-reply reminder; subsequent messages carry it with probability
-  // REMINDER_PROBABILITY. Historical replay does not count.
-  let inboundMessageCount = 0
-  const REPLY_REMINDER = 'Substantive replies should be posted to IRC.'
-  const REMINDER_PROBABILITY = 1 / 7
+  // Tracks whether any non-historical inbound message has been emitted in
+  // this session — gates the always-attach-on-first-message behavior.
+  let firstMessageSeen = false
 
   // ---- MCP server --------------------------------------------------------
 
@@ -194,11 +196,11 @@ export function createMcpServer(client: RoostIrcClient, config: ClientConfig): {
 
     let content = msg.text
     if (!meta.historical) {
-      inboundMessageCount++
-      if (inboundMessageCount === 1 || Math.random() < REMINDER_PROBABILITY) {
+      if (!firstMessageSeen || Math.random() < REMINDER_PROBABILITY) {
         content = `${msg.text}\n\n${REPLY_REMINDER}`
         metaRecord.reminder = 'true'
       }
+      firstMessageSeen = true
     }
 
     pushNotification(content, metaRecord)
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -194,19 +194,23 @@ export function createMcpServer(client: RoostIrcClient, config: ClientConfig): {
     }
     if (meta.historical) metaRecord.historical = 'true'
 
-    let content = msg.text
-    if (!meta.historical) {
-      if (!firstMessageSeen || Math.random() < REMINDER_PROBABILITY) {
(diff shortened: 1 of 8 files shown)


src/irc-server.ts
```diff
@@ -184,7 +193,17 @@ export function createMcpServer(client: RoostIrcClient, config: ClientConfig): {
       if (meta.chunkCount && meta.chunkCount > 1) metaRecord.chunkCount = String(meta.chunkCount)
     }
     if (meta.historical) metaRecord.historical = 'true'
-    pushNotification(msg.text, metaRecord)
+
+    let content = msg.text
+    if (!meta.historical) {
+      if (!firstMessageSeen || Math.random() < REMINDER_PROBABILITY) {
+        content = `${msg.text}\n\n${REPLY_REMINDER}`
```
AlexSc: I want the reminder to be clearly separate from the message content.

AlexSc (changes requested):

src/irc-server.ts
AlexSc: Suggestion: push it as an immediate followup system message

TeakBuilds: [worker-136] proposing this shape — keeps it in the same notification (per the issue spec) but makes the boundary unambiguous via a tagged line. Concretely:

```
<message text>

[reminder] Substantive replies should be posted to IRC.
```

i.e. `\n\n[reminder] Substantive replies should be posted to IRC.` appended after the message body. The `[reminder]` prefix is a single-purpose marker so agents can visually parse it as out-of-band guidance, not a continuation of the message. Open to alternatives — divider line, distinct prefix, etc. — if you have a specific shape in mind.

TeakBuilds: [worker-136] ack — implementing as immediate followup notification with `event=reminder`. Reverting the content-mutation path and the test predicate relaxations that path required.

test/irc-server.test.ts
```diff
@@ -239,6 +239,124 @@ describe.if(isErgoAvailable())('irc-server MCP tools', () => {
...
+    expect(Number(reminder.meta.seq)).toBeGreaterThan(Number(msg.meta.seq))
+  })
+
+  it('historical replay does not emit a reminder notification', async () => {
+    const peer = await connectPeer(ergo, 'ip-rem3-peer')
+    const mcp = await startMcpInProcess(ergo, 'ip-rem3')
+
+    await peer.joinChannel('#ip-rem3-backfill')
+    peer.say('#ip-rem3-backfill', 'historical-1')
+    await sleep(200)
```
AlexSc: I would rather not sleep, the test suite is already taking a surprisingly long time to run for what we do.

test/irc-server.test.ts
```diff
@@ -88,7 +88,7 @@ describe.if(isErgoAvailable())('irc-server MCP tools', () => {
 
     peer.say('#ip-inbound1', 'ping from peer')
     const n = await mcp.waitForNotification(
-      n => n.meta.channel === '#ip-inbound1' && n.content === 'ping from peer',
+      n => n.meta.channel === '#ip-inbound1' && n.content.startsWith('ping from peer'),
```
AlexSc: Do we still need startsWith checks?

AlexSc (changes requested):
Good work on the followup reply, just some test cleanup I wan to see land

test/irc-server.test.ts
TeakBuilds: [worker-136] all `startsWith` predicates in test/irc-server.test.ts reverted to `===` — msg.text is no longer mutated so exact match works.

test/irc-server.test.ts
TeakBuilds: [worker-136] dropped the sleeps. Replaced the historical-setup `sleep(200)` with peer self-echo wait (enabled `echo-message` cap on the peer helper) — guarantees ergo has stored the message before mcp joins. Replaced the pre-assertion `sleep(100)` with event-driven fences (peer leave → membership notif). Also folded the redundant "no reminder for historical" test into the slot test as a single-reminder count assertion.

AlexSc (approved):

PR #171, opened 2026-05-07
fix: prevent bun runner hang from abandoned helper rejections (#170)

### LINKED ISSUE #170: fix: bun test runner hangs indefinitely after a startMcp subprocess test fails
**Symptom:** when a single test using \`startMcp\` (subprocess MCP test runner — \`test/helpers/mcp.ts\`) fails or times out, bun's test runner doesn't exit. It logs \`killed N dangling process(es)\` and then hangs indefinitely. CI on PR #169 hung for 58 minutes until GHA cancellation; locally same.

**Repro** (on PR #169 branch, but any subprocess-test failure should reproduce):
\`\`\`
timeout 90 bun test test/chathistory.test.ts
# exit 124 (timeout) — bun never returned
\`\`\`

The specific test that triggers the cascade is \`chathistory backfill (subprocess) > SIGUSR1 clears seen-set so messages backfill after compaction reset\`, but the test content isn't the issue — any subprocess test that fails to clean up causes the runner to wedge.

**Why it matters:** every uncaught test failure in a subprocess test class becomes a 58-minute CI burn. We just hit it on the alpha milestone work and the diagnosis ate hours of agent and human time. Without this fix, any future bad test takes the runner down.

**Investigation surface:**
- \`test/helpers/mcp.ts\` (\`startMcp\`) — how does it spawn? What handles does it leak on test failure?
- Does the spawned MCP keep an open stdio pipe / socket that the bun parent waits on?
- Is the bun runner's "kill dangling process" path actually killing it, or just losing track?
- Does \`afterAll\` / \`afterEach\` reliably tear down on timeout? Bun's behavior here may differ from jest/vitest.
- Compare with \`startMcpInProcess\` (in-process variant) which doesn't have this issue.

**Acceptance:**
- A subprocess test that times out or throws causes the test runner to exit cleanly (non-zero) within reasonable time (≤30s).
- No \`killed N dangling process\` messages, or if there are, the runner still terminates.
- Add a regression test if feasible: a deliberately-hanging subprocess test wrapped in a way that proves the cleanup path works.

Discovered while debugging PR #169 timeouts. See \`/test-timeouts-debug\` channel session and lead-pm summary in \`#leads-roost-dev\`.

TeakBuilds:
## Summary
- Bun 1.2.20 deadlocks the test runner when an unhandled promise rejection arrives from a test the runner has already abandoned (timeout, prior failure).
- Our wait-style helpers (`waitForNotification`, `waitForMessage`, `waitForPart`, `joinChannel`, `changeNick`) race the user's `await` against bun's test timeout via `setTimeout(reject, ...)` — when bun aborts first, the helper's `reject()` fires later on a Promise nobody is awaiting, becomes unhandled, and wedges the runner. That's the 58-minute CI burn from PR #169.
- Fix: pre-attach a no-op `.catch()` (`suppressLateRejection`) to each helper's inner Promise. A still-active `await` keeps throwing normally; an abandoned rejection is silently absorbed.
- Adds a regression test that exercises the abandoned-rejection path and a CLAUDE.md note documenting the footgun.

Closes #170.

## Followups (lead-pm to triage)
- File upstream bun bug. Minimal repro from the diagnosis (no roost code at all):
  ```ts
  import { describe, it } from 'bun:test'
  describe('repro', () => {
    it('test 1', async () => {
      new Promise((_, reject) => { setTimeout(() => reject(new Error('boom')), 200) })
      await new Promise(() => {})
    })
    it('test 2', async () => { await new Promise(r => setTimeout(r, 100)) })
  })
  ```
  Bun 1.2.20 hangs indefinitely on this — test 1 fails at 200ms via the unhandled rejection, then bun never proceeds to test 2 and never exits.

## Test plan
- [x] `bun test test/abandoned-rejection.test.ts` — passes (regression test)
- [x] `bun test` full suite — 184 pass, 0 fail, ~65s
- [x] Hang verification: induced a deliberate `waitForNotification` predicate-never-matches failure in `chathistory.test.ts` subprocess test; without the fix it hangs indefinitely, with the fix it exits cleanly in ~8s with exit code 1.
- [ ] CI runs full suite

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,6 +21,8 @@ Requires ergo (IRCv3 server). Install with `bin/install-ergo` or set `ERGO_BIN` 
 
 **Never pipe `bun test` through `tail`, `head`, or `grep`.** The shell returns the *last* command's exit code, which is always 0 for those filters — so `bun test ... | tail -N` reports success even when bun failed. The pipe also buffers the entire run until exit, masking hangs. If you need to trim noisy output, write the full result to a file and read it: `bun test ... > /tmp/out 2>&1; echo "exit=$?"; tail -N /tmp/out`. The exit code is the only honest signal.
 
+**Bun-specific footgun (issue #170, upstream report TBD):** when an unhandled promise rejection arrives from a test the runner has already abandoned, bun 1.2.20 deadlocks the runner (no subsequent test runs, process never exits). Any test helper that resolves/rejects via a `setTimeout(reject, ...)` race against the user's `await` will hit this if the user's await is ever aborted (test timeout, prior failure, etc.). Wrap such promises with `suppressLateRejection` from `test/helpers/tool.ts`: it pre-attaches a no-op `.catch()` so an abandoned rejection is silently absorbed, while a still-active `await` still throws normally. The existing wait-style helpers in `test/helpers/mcp-core.ts` and `test/helpers/peer.ts` already use it.
+
 ## Worktrees
 
 Use `script/worktree <branch> [--from <base>] [path]` to bootstrap a new worktree — it creates the sibling worktree, runs `bun install`, and copies `.claude/settings.local.json` from the main worktree so spawned workers don't hit a permission-prompt flood.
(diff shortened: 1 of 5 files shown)


AlexSc (approved):
Good work tracking down a library bug.

PR #176, opened 2026-05-07
fix: narrow pushNotification catch — only suppress transport-closed errors

### LINKED ISSUE #129: fix: mcp.notification rejection catch swallows real errors (L7)
`mcp.notification(...).catch(() => {})` at irc-server.ts:217 swallows any rejection. The comment names "transport teardown" as the intended case, but real bugs would also be silently swallowed.

Fix: narrow the catch — check the error message for the known transport-closed shape; anything else gets a `process.stderr.write`.

**Acceptance:**
- Catch only suppresses errors matching the transport-closed pattern
- Other rejections produce a stderr log entry

TeakBuilds:
Fixes #129.

`pushNotification` now only swallows rejections whose message matches `/not connected/i` (the MCP SDK's transport-teardown shape). Anything else gets a `process.stderr.write`.

Two unit tests added (no ergo required): one for the silent-suppress path, one for the stderr-log path.

### DIFF
--- a/test/irc-server.test.ts
+++ b/test/irc-server.test.ts
@@ -8,7 +8,7 @@ import { startErgo, isErgoAvailable, type ErgoContext } from './helpers/ergo.js'
 import { startMcpInProcess } from './helpers/mcp-inprocess.js'
 import { startMcp } from './helpers/mcp.js'
 import { connectPeer } from './helpers/peer.js'
-import { toolText, sleep } from './helpers/tool.js'
+import { toolText } from './helpers/tool.js'
 import { createMcpServer } from '../src/irc-server.js'
 import type { RoostIrcClient, ClientConfig } from '../src/irc-client.js'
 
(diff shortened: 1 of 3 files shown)


test/irc-server.test.ts
```diff
@@ -465,3 +468,67 @@ describe.if(isErgoAvailable())('irc-server MCP tools (subprocess)', () => {
...
+}
+
+describe('pushNotification error handling', () => {
+  it('silently suppresses Not connected (transport teardown)', async () => {
+    const { emitUnreadSummary } = createMcpServer(makeStubClient(), stubConfig)
+
+    const stderrSpy = spyOn(process.stderr, 'write').mockImplementation((() => true) as typeof process.stderr.write)
+    try {
+      emitUnreadSummary()
+      await sleep(50)
```
AlexSc: I'd really rather not sleep

AlexSc (changes requested):

AlexSc (changes requested):
I'm actually just debugging a dispatcher issue. You can re-request review.

AlexSc (approved):

PR #183, opened 2026-05-07
feat: channel_join returns NAMES list (#113)

### LINKED ISSUE #113: MCP: channel_join should return current NAMES (and ideally history pointer) in tool result
Currently `channel_join` returns just "joined #foo" — no NAMES list, no history pointer. To know who's in a channel I just joined I have to call `channel_who` separately, and recent history arrives as a separate stream of <channel> notifications rather than as part of the join return value.

Felt awkward joining #issue-111 with no idea who else was there until an extra round trip.

Proposal: `channel_join` returns `{ joined: true, members: [...], historyOldestSeq?: N }` (or similar). Saves a round trip, matches user mental model — when you walk into a room you can see who's there.

Open question: should history come inline in the return value too, or stay as the streamed-notification model? Inline is simpler for the agent but bloats the join return on busy channels. Probably fine to leave history as-is and just add NAMES.

Filed by @lead-pm during alpha cleanup discussion.

TeakBuilds:
closes #113

`channel_join` now waits for NAMES before resolving and returns the member list inline.

**shape:** `joined #foo\nmembers (N): nick1, nick2, ...`

**how it works:** irc-framework coalesces all RPL_NAMEREPLY (353) into a cache and fires `userlist` exactly once on RPL_ENDOFNAMES (366). `handleJoin` defers resolution until `handleUserlist` fires. 2s NAMES timeout guards against servers that stall on NAMES — resolves with `{ ok: true, members: [self] }` in that case. Original 5s join-ack timeout still guards the JOIN itself.

**tests added:** multi-user channel returns full list, solo channel returns just self, NAMES-timeout path resolves with self only.

[worker-113]

### DIFF
--- a/test/disconnect-cache.test.ts
+++ b/test/disconnect-cache.test.ts
@@ -149,11 +149,11 @@ describe('draft/multiline cap malformed value warning', () => {
 describe('socket close pre-empts pending join/part resolvers', () => {
   it('join resolver resolves false immediately on socket close', async () => {
     const client = makeClient()
-    const p = new Promise<boolean>(resolve => {
+    const p = new Promise<{ ok: boolean; members: string[] }>(resolve => {
       client.joinResolvers.set('#chan', [resolve])
     })
     client.handleSocketClose()
-    expect(await p).toBe(false)
+    expect((await p).ok).toBe(false)
   })
 
   it('part resolver resolves false immediately on socket close', async () => {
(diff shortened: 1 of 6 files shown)


test/disconnect-cache.test.ts
```diff
@@ -149,11 +149,11 @@ describe('draft/multiline cap malformed value warning', () => {
 describe('socket close pre-empts pending join/part resolvers', () => {
   it('join resolver resolves false immediately on socket close', async () => {
     const client = makeClient()
-    const p = new Promise<boolean>(resolve => {
+    const p = new Promise<import('../src/irc-client.js').JoinResult>(resolve => {
```
AlexSc: That is an insanely ugly import

AlexSc (changes requested):
Fix the ridiculous imports

AlexSc (approved):

PR #184, opened 2026-05-07
fix: single source of truth for IRC loopback host

### LINKED ISSUE #144: drop hardcoded 127.0.0.1 loopback in favor of a single source
@alex flagged on #141: 127.0.0.1 is hardcoded in multiple places. Right now:

- \`bin/roost:98\` — error message
- \`bin/roost:308\` — \`ROOST_PERM_HOST=127.0.0.1\` injected into worker tmux env
- \`bin/roost:436\` — startup message
- \`bin/roost-permbot:33\` (docstring) and \`:57\` (\`os.environ.get('ROOST_PERM_HOST', '127.0.0.1')\`)
- \`src/irc-server.ts:16\` (docstring) and \`:853\` (\`env('ROOST_IRC_SERVER', '127.0.0.1')\`)

Two consumer chains (roost CLI / permbot, and the MCP server) each carry their own default. They'll drift.

Goal: single source of truth for the IRC host — probably an env-with-default in \`bin/roost\` that everything downstream reads. Permbot and irc-server both already accept \`ROOST_PERM_HOST\` / \`ROOST_IRC_SERVER\`; \`bin/roost\` should set whichever is missing instead of duplicating the literal.

Tagged beta — alpha is functional today on loopback.

TeakBuilds:
Closes #144.

`bin/roost` now reads `ROOST_IRC_SERVER` (defaulting to `127.0.0.1`) into a local `IRC_HOST` variable and uses it everywhere — error messages, startup output, and both child-process env injections (`ROOST_IRC_SERVER` for the MCP, `ROOST_PERM_HOST` for permbot). Previously the MCP server fell back to its own hardcoded default with no connection to what `bin/roost` was using.

[worker-144] 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -16,6 +16,7 @@
 set -uo pipefail
 
 ROOST_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+IRC_HOST="${ROOST_IRC_SERVER:-127.0.0.1}"
 IRC_PORT=6667
 SESSION_PREFIX="roost-"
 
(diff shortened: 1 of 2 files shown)


AlexSc:
Follow-ups are exhausting just fix it here please

AlexSc:
What's the actual behavior of the system when IRC isn't up where expected? I've used systems without `nc`, slightly concerned about unexpected blow up in that case

AlexSc (approved):

PR #186, opened 2026-05-08
feat: prompts/ template library + --prompt-template flag

### LINKED ISSUE #85: prompts/: agent prompt template library
[lead-pm] surfaced live during alpha runs — every worker/reviewer/watcher spawn re-types a longhand prompt, and the prompts drift in shape between launches. Worth baking templates into the repo.

## Shape

- \`prompts/worker.md\` — base worker prompt with \`\${ISSUE}\`, \`\${BRANCH}\` placeholders
- \`prompts/reviewer.md\` — reviewer/simplify prompt with \`\${PR}\`, \`\${PR_DESCRIPTION}\` placeholders
- \`prompts/watcher.md\` — the haiku-overseer prompt (already shared on #67)
- \`prompts/lead-pm.md\` — the start.md content, normalized

\`bin/roost\` could grow a \`--prompt-template <name>\` flag that loads from \`prompts/\${name}.md\` and substitutes any env-var-style placeholders from CLI flags. e.g.:

    bin/roost spawn worker-65 -c '#issue-65' --prompt-template worker \\
      --prompt-arg ISSUE=65 --prompt-arg BRANCH=maint/issue-65-tests-inprocess

Or simpler: just \`bun template prompts/worker.md ISSUE=65 BRANCH=...\` and pipe to \`--prompt-stdin\`.

## Why beta

Real ergonomics win for any orchestrator class agent (lead-pm, future product PMs, etc.). Not blocking alpha.

TeakBuilds:
closes #85

Adds `prompts/{worker,reviewer,watcher,lead-pm}.md` and the `--prompt-template NAME` + `--prompt-arg KEY=VALUE` flags on `roost spawn`.

The four templates are lifted from prompts I've been hand-typing this session — worker, reviewer, watcher (just wraps `/watcher`), and lead-pm (the parameterized start.md).

Substitution is dumb bash parameter expansion (`${var//\${KEY}/$VAL}`). No regex, no `eval`, no external tool. Unsubstituted placeholders, missing template files, malformed `KEY=VALUE` args, and combining `--prompt-template` with `--prompt`/`--prompt-stdin`/`--prompt-file` all fail loudly.

5 new tests cover the error paths. Manual smoke verified the success path. shellcheck clean, full test suite green (199 ran, 126 pass, 73 skip from no-ergo, 0 fail).

Out of scope:
- Migrating start.md → prompts/lead-pm.md (kept both for now; lead-pm.md references the parameterized version)
- Test coverage for the success-path substitution (would require a dry-run mode or env-var test hook)
- Auto-resolving REPO/BRANCH from git context — caller passes them explicitly


AlexSc (changes requested):
Because the prompt templates are so deeply tied to what we do here, I wouldn't bake them into roost spawn -- inappropriate for other users.

I would leave them in prompts as examples, but then symlink to skills/commands like watcher currently is. Gets us "real world examples" to reference in docs, and you get simple slash commands to use as prompts. Thoughts?

prompts/worker.md
```diff
@@ -0,0 +1,18 @@
+---
+description: Roost worker — implements an issue on a feature branch, drafts a PR, defers to lead-pm for ready/review/cleanup.
+argument-hint: [issue-number] [owner/repo] [branch-name]
+---
+You are worker-$0 on Roost (an IRC-mediated agent harness). You're in #issue-$0 with @lead-pm (your project lead) and @alex (the human). The channel is the authoritative source of input — alex will not message you directly after spawn, only via the channel.
+
+Your task: GitHub issue $1#$0. Read it yourself with gh. Branch `$2` is checked out here.
+
+Process:
+1. Read the issue and any relevant code
```
AlexSc: Would like to push it to use the GitHub management skill, which includes additional data. At the very least we need to be sure it gets comments

AlexSc (changes requested):

prompts/worker.md
TeakBuilds: [lead-pm] Pushed: worker step 1 now points to `scripts/view-issue.sh` from the github-management skill, with a note on why plain `gh` is insufficient.

AlexSc (approved):

PR #190, opened 2026-05-08
permbot: collapse into MCP process + owner-gate via session id

### LINKED ISSUE #188: permbot: silent disconnect blocks worker indefinitely
[lead-pm] surfaced live during #137 work.

The permbot side daemon (started by `--perm-irc` in `roost spawn`) lost its IRC connection mid-session and didn't reconnect. The parent worker hit a permission-prompt hook, sent the prompt to a no-longer-listening permbot, and blocked on the timeout fallback. Each subsequent prompt required a manual `roost send worker-137 y` to push input directly into the worker's tmux pane.

## Repro (live observation)

worker-137 was running with `--perm-irc --perm-target lead-pm`. After ~hours of normal operation:
- last permbot DM to lead-pm at 23:19:41 (a normal y/n prompt)
- silence after that — permbot never DMed again
- worker would hang on `Bash` calls; only `roost send worker-137 y` unstuck it
- `permbot-worker-137` accumulated 35+ unread DMs from the *worker* asking it to relay, with no responses

## Why it matters

When permbot dies silently the worker becomes operator-bound — every tool call needs manual unblocking via tmux. Defeats the point of `--perm-irc`. Also a safety regression: the operator loses the structured y/n channel and falls back to typing into a pane.

## Fix shape

Two layers:

1. **Detect**: permbot should heartbeat (e.g. log "alive" every interval, or expose a status endpoint). Right now you only notice it's dead when prompts stop arriving.
2. **Recover**: on disconnect, permbot should auto-reconnect (the worker's MCP IRC client already does this via `irc-framework`). Could also have the prompt hook attempt to start permbot on demand when no recipient is reachable.

(2) is the durable fix. Pairs with #82 (stale intent display) — both about making the permission-prompt path reliable enough to trust.

## Why beta

Real reliability bug observed live. Doesn't block alpha (worktree workflow still functions with manual unblocking) but undermines the `--perm-irc` value prop.

## Out of scope

- Replacing the prompt-hook architecture wholesale
- Multi-permbot-per-worker

TeakBuilds:
## Summary

- **Collapse permbot into the MCP process.** `src/permbot.ts` becomes a module loaded by `src/irc-server.ts`; the MCP now holds two `RoostIrcClientImpl`s (worker + `permbot-${NICK}`). Drops the `Bun.spawn(permbot.ts)` block, the standalone `bin/roost-permbot` launcher, and the parent-reparent watcher.
- **Owner gate via `CLAUDE_CODE_SESSION_ID`.** New `src/owner-gate.ts`. First MCP per `ROOST_DATA_DIR` atomic-creates `owner.session` (`O_EXCL`); later starters with a mismatching session id stay passive (no IRC nick claim, no permbot, no socket bind, no pidfile write — tools return `PASSIVE_SENTINEL`). The `irc-permission-prompt` hook does the same compare and short-circuits to `emit('ask', ...)` BEFORE any socket touch, so nested claudes get the normal terminal prompt instead of DMing the parent's operator.

closes #188.

## Why

Root cause from #188 (lead-pm + alex): nested claudes inherit the worker's `ROOST_DATA_DIR`/`ROOST_PERM_SOCK`/`ROOST_IRC_NICK` via tmux env. The user-installed roost plugin in the nested claude auto-spawned a second MCP, which spawned a second permbot with the same `permbot-${NICK}` — ergo kicked the original off via nick collision and the worker was permbot-less for the rest of the session. Heartbeat/reconnect treats the symptom; the real fix is making duplicate MCPs benign.

## Test plan

- [x] `bun test` (216 pass, 0 fail; ergo-driven integration tests pass)
- [x] `bun tsc --noEmit` clean
- [x] new `test/owner-gate.test.ts` — claim/check matrix incl. EEXIST-with-different-session goes passive
- [x] new case in `test/permission-prompt.test.ts` — owner-gate short-circuits hook BEFORE binding a unix listener (asserts no client connection observed on the listener)
- [x] existing `test/permbot.test.ts` unchanged and still passes (it always tested `startPermbot()` against a mock client — that's the public surface)

Drive-by: `test/helpers/mcp.ts` now scrubs `ROOST_*` env vars from the spawned subprocess so test runs from inside a live worker session don't accidentally bind a test permbot on top of the operator's real socket. Pre-existing leak; surfaced once permbot ran in-process and logged through the same nick.

## Notes for review

- Open Q from plan: claude's behavior on a crashed MCP (auto-restart? loop?) is unverified — defaulted to passive-stay-alive per lead-pm's pressure. If the next nested-claude scenario shows passive is overkill, can switch to exit-at-startup.
- `bin/roost-permbot` deleted (no back-compat per CLAUDE.md).
- audit doc (`docs/audit-2026-05-mcp.md`) section "Findings — bin/roost-permbot" gets a "resolved by #188" note rather than rewriting; the line refs were already stale.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,10 @@ When a Claude Code session loads `roost-irc` as an MCP and connects:
 /plugin install roost@roost
 ```
 
-This puts `roost`, `roost-permbot`, and `irc-permission-prompt` on
-your PATH and auto-loads the `roost-irc` MCP for every session.
+This puts `roost` and `irc-permission-prompt` on your PATH and
+auto-loads the `roost-irc` MCP for every session. The permbot
+routing daemon (used by `--perm-irc`) runs in-process inside the
+MCP — no separate launcher.
 
 After installing, pull dependencies:
 
@@ -137,9 +139,9 @@ irssi -c 127.0.0.1 -n myname
 
 ## IRC permission oversight (--perm-irc)
 
-`--perm-irc` starts a side daemon (`roost-permbot`) alongside the
-worker. The daemon holds a stable IRC nick `permbot-{worker}` and
-serializes the worker's PreToolUse permission prompts as DMs to
+`--perm-irc` runs a permbot routing module inside the worker's MCP
+process. The module holds a second IRC connection on a stable nick
+`permbot-{worker}` and serializes the worker's PreToolUse permission prompts as DMs to
 `--perm-target` (required). The operator replies `y` / `n` / `yes` /
 `no` / `allow` / `deny`; anything else or a 30s timeout falls through
 to the terminal prompt.
@@ -225,7 +227,6 @@ roost/
 ├── bin/
 │   ├── roost               Wrapper: spawn / shutdown / list / attach / tail / status / root.
 │   ├── roost-irc-server    PATH-resolvable launcher for the MCP server.
-│   ├── roost-permbot       IRC permission oversight daemon.
 │   └── irc-permission-prompt  PermissionRequest hook (thin socket client, loaded per-session by --perm-irc).
 ├── etc/ergo.yaml           Sample ergo IRC server config.
 ├── extras/weechat/         Optional weechat notification script.
--- a/README.md
+++ b/README.md
@@ -44,9 +44,7 @@ When a Claude Code session loads `roost-irc` as an MCP and connects:
 ```
 
 This puts `roost` and `irc-permission-prompt` on your PATH and
-auto-loads the `roost-irc` MCP for every session. The permbot
-routing daemon (used by `--perm-irc`) runs in-process inside the
-MCP — no separate launcher.
+auto-loads the `roost-irc` MCP for every session.
 
 After installing, pull dependencies:
 
@@ -141,7 +139,7 @@ irssi -c 127.0.0.1 -n myname
 
 `--perm-irc` runs a permbot routing module inside the worker's MCP
 process. The module holds a second IRC connection on a stable nick
-`permbot-{worker}` and serializes the worker's PreToolUse permission prompts as DMs to
+`permbot-{worker}` and serializes the worker's PermissionRequest prompts as DMs to
 `--perm-target` (required). The operator replies `y` / `n` / `yes` /
 `no` / `allow` / `deny`; anything else or a 30s timeout falls through
 to the terminal prompt.
--- a/skills/roost/SKILL.md
+++ b/skills/roost/SKILL.md
@@ -61,13 +61,14 @@ claude flag the wrapper doesn't otherwise know about.
 
 ## IRC permission oversight (--perm-irc)
 
-`--perm-irc` starts a side daemon (`bin/roost-permbot`) alongside the
-worker. The daemon holds a stable IRC nick `permbot-{worker}` and
-serializes the worker's PreToolUse permission prompts as DMs to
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -43,8 +43,10 @@ When a Claude Code session loads `roost-irc` as an MCP and connects:
 /plugin install roost@roost
 ```
 
-This puts `roost`, `roost-permbot`, and `irc-permission-prompt` on
-your PATH and auto-loads the `roost-irc` MCP for every session.
+This puts `roost` and `irc-permission-prompt` on your PATH and
```
AlexSc, line 46: We don't need historical breadcrumbs.

README.md
```diff
@@ -137,9 +139,9 @@ irssi -c 127.0.0.1 -n myname
 
 ## IRC permission oversight (--perm-irc)
 
-`--perm-irc` starts a side daemon (`roost-permbot`) alongside the
-worker. The daemon holds a stable IRC nick `permbot-{worker}` and
-serializes the worker's PreToolUse permission prompts as DMs to
+`--perm-irc` runs a permbot routing module inside the worker's MCP
+process. The module holds a second IRC connection on a stable nick
+`permbot-{worker}` and serializes the worker's PreToolUse permission prompts as DMs to
```
AlexSc: PermissionRequest prompts

skills/roost/SKILL.md
```diff
@@ -61,13 +61,14 @@ claude flag the wrapper doesn't otherwise know about.
...
-`--perm-irc` starts a side daemon (`bin/roost-permbot`) alongside the
-worker. The daemon holds a stable IRC nick `permbot-{worker}` and
-serializes the worker's PreToolUse permission prompts as DMs to
-`--perm-target` (required). The operator replies `y` / `n` /
-`yes` / `no` / `allow` / `deny` (case-insensitive); anything else or a
-30s timeout falls through to the regular terminal prompt as a safety
-net. `roost shutdown` reaps the daemon and its socket/pidfile.
+`--perm-irc` runs a permbot routing module inside the worker's MCP
+process. It opens a second IRC connection on the stable nick
+`permbot-{worker}` and serializes the worker's PreToolUse permission
```
AlexSc: PermissionRequest prompts

skills/roost/SKILL.md
```diff
@@ -61,13 +61,14 @@ claude flag the wrapper doesn't otherwise know about.
...
+prompts as DMs to `--perm-target` (required). The operator replies
+`y` / `n` / `yes` / `no` / `allow` / `deny` (case-insensitive); anything
+else or a 30s timeout falls through to the regular terminal prompt as
```
AlexSc: Cut timeout mention

AlexSc (changes requested):

README.md
TeakBuilds: [worker-188] fixed in 9e6988b

README.md
TeakBuilds: [worker-188] fixed in 9e6988b

skills/roost/SKILL.md
TeakBuilds: [worker-188] fixed in 9e6988b

skills/roost/SKILL.md
TeakBuilds: [worker-188] fixed in 9e6988b

AlexSc (approved):

PR #191, opened 2026-05-08
docs: clarify MCP is a plain IRC client, no dispatcher pipeline

### LINKED ISSUE #74: MCP docs pass: review for agent comprehension (esp. dispatcher ↔ MCP boundary)
The MCP documentation could use a pass focused on what an agent (not a human reading carefully) is likely to internalize on a first read.

## Concrete miss

In a recent lead-pm session, I claimed: \"lead-pm sees IRC via the MCP, not via the dispatcher pipeline, so multiline doesn't help me — only fixes the worker-facing channels.\"

That's wrong. The MCP is just an IRC client. The dispatcher posts to whatever channels it routes to, including the project channel the lead-pm joined. Full multiline relays land in lead-pm's channel exactly the same way they land in worker channels. I lost the thread on \"this is just your IRC client\" and constructed a separate \"MCP receives from dispatcher\" mental model that doesn't exist.

## Ask

Audit `README`, `docs/ORCHESTRATOR.md`, and any MCP-tool inline help with this lens:

- Does the doc make it obvious that the MCP is a plain IRCv3 client, and that channels/messages have no special routing layer between dispatcher and MCP?
- Are there places where \"MCP\" and \"dispatcher\" sound coupled when they're not?
- Could a fresh agent read the docs once and avoid the mental model I built?

## Out of scope

- Rewriting docs from scratch
- Adding architecture diagrams (probably worth it but separate issue if so)

[lead-pm]

TeakBuilds:
fixes #74

four targeted sentences across `src/irc-server.ts`, `README.md`, `docs/ORCHESTRATOR.md`, and `ARCHITECTURE.md`. no rewrites.

the `instructions` string in `irc-server.ts` leads — every agent reads it at MCP boot before any README. it now opens with: *"This MCP is a plain IRCv3 client — there is no special pipeline between it and any other component. Every message that arrives in a channel (from another agent, a human, or the dispatcher script) reaches you identically, as a normal IRC channel message."*

the other three files reinforce that same point in context.

🤖 Generated with Claude Code


AlexSc (changes requested):
Worth a quick note in the mcp instructions that multiline is permitted. Cut dispatcher reference, just say "bots" instead

AlexSc (approved):

PR #199, opened 2026-05-08
fix: waitForMessage timeout cleanup never found the waiter (#172)

### LINKED ISSUE #172: test/helpers: waitForMessage timeout-cleanup never finds the waiter
**Pre-existing**, surfaced by review of #171.

In \`test/helpers/peer.ts\` \`waitForMessage\`, the timeout cleanup does:
\`\`\`ts
const idx = messageWaiters.findIndex(w => w.resolve === resolve)
\`\`\`
…but the entry pushed into \`messageWaiters\` stores a *wrapped* resolve:
\`\`\`ts
messageWaiters.push({
  channel,
  pred,
  resolve: (msg) => { clearTimeout(timer); resolve(msg) },
  reject,
})
\`\`\`
\`findIndex\` compares the wrapped function against the raw \`resolve\`, so it always returns -1. The waiter is never removed on timeout — it leaks until a future message happens to satisfy the predicate.

Probably benign in practice (each test uses fresh peers, and the bun footgun fix from #170 keeps the late rejection from wedging the runner). Worth fixing for hygiene: store/compare against the same reference.

Discovered by reviewer-170 during PR #171 review. Not in scope of #170 fix.

TeakBuilds:
closes #172

`findIndex` compared against raw `resolve` but the pushed entry stored a wrapped function — always -1, so the waiter leaked on timeout. Fix: extract `wrappedResolve`, use it in both the push and the findIndex.

Also adds `pendingWaiterCount` to `PeerContext` and two regression tests in `helpers.test.ts` (timeout path + success path).

218/218 tests pass.

### DIFF
--- a/test/helpers.test.ts
+++ b/test/helpers.test.ts
@@ -41,7 +41,7 @@ describe.if(isErgoAvailable())('test helpers', () => {
     expect(mcp.waiterCount()).toBe(0)
   })
 
-  it('waitForMessage removes waiter on timeout — issue #172 regression', async () => {
+  it('waitForMessage removes waiter on timeout', async () => {
     const peer = await connectPeer(ergo, 'wfm-timeout-peer')
     await peer.joinChannel('#wfm-timeout')
     await expect(peer.waitForMessage('#wfm-timeout', () => false, 50)).rejects.toThrow('timed out')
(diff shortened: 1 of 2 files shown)


test/helpers.test.ts
```diff
@@ -41,6 +41,24 @@ describe.if(isErgoAvailable())('test helpers', () => {
     expect(mcp.waiterCount()).toBe(0)
   })
 
+  it('waitForMessage removes waiter on timeout — issue #172 regression', async () => {
```
AlexSc: Prefer timeless test name

AlexSc (changes requested):

AlexSc (approved):

PR #203, opened 2026-05-09
feat: roost init — bootstrap .orchestrator/config.json

### LINKED ISSUE #202: roost init: bootstrap .orchestrator/config.json for a new project
## Goal

`roost init` is the once-per-project command users run to bootstrap roost in their codebase. It writes `.orchestrator/config.json` with the minimum config needed for the dispatcher + lead-pm spawn to work, autodetecting everything it can.

This is step 2 of the four-command get-started flow (#195):

```
brew install AlexSc/tap/roost
cd ~/Dev/myproject
roost init
roost spawn myproject-lead-pm --channels '#myproject-leads' --prompt '/lead-pm myproject beta alex AlexSc'
```

Implemented as a bash subcommand of `bin/roost`. Pure arg-parse + autodetection. No prompts. No new deps.

## What it writes

`.orchestrator/config.json` matching `OrchestratorConfig` in `src/orchestrator/config.ts`:

```json
{
  "project": "myproject",
  "repo": "Owner/repo",
  "agent_logins": ["alexsc"],
  "watched_issues": [],
  "watched_prs": []
}
```

Defaults from `src/orchestrator/naming.ts` cover `irc.nick`, `irc.project_channel`, `irc.server`, `irc.port`, `irc.interval_seconds` — `roost init` does not write them. Keeps the on-disk config minimal and diffs from future schema changes easy to read.

Also writes `.orchestrator/.gitignore` with:

```
state.json
last-tick.txt
last-error.txt
```

These are dispatcher-owned, churn frequently, and shouldn't be committed.

## Inputs

Each value is autodetected; the corresponding flag overrides the autodetect.

| Field          | Autodetect                                              | Flag             |
|----------------|---------------------------------------------------------|------------------|
| `project`      | `basename(cwd)` lowercased                              | `--project`      |
| `repo`         | `git remote get-url origin` → parse `Owner/repo`        | `--repo`         |
| `agent_logins` | `gh api user --jq .login` (single-element list default) | `--agent-login` (repeatable) |

If autodetection fails for a required field and no override flag was given, exit non-zero with a clear hint (e.g. "no `origin` remote and no `--repo` flag — pass `--repo Owner/repo` or run `git remote add origin ...`").

`irc.server`, `irc.port`, `irc.interval_seconds` are not flagged — defaults are enough for the local-ergo-on-127.0.0.1 case, which is the dogfood path. Users who need non-default IRC config edit config.json by hand.

## Preconditions

Hard-error with actionable hints if any of these fail:

- cwd is not a git repo (`git rev-parse --is-inside-work-tree` fails) — "run `git init` first or `cd` to a git repo"
- `gh auth status` is not authed — "run `gh auth login` first"
- the resolved `repo` is not reachable via `gh repo view` — "check the repo name + your gh auth"
- the project name does not match `^[a-z0-9][a-z0-9-]*$` (mirrors `validateProject` in naming.ts) — exit with the validator's error message

## Existing-config behavior

If `.orchestrator/config.json` already exists, `roost init` exits with a "already initialized" message and a non-zero status. Pass `--force` to overwrite without merging — explicitly **no migration logic**, even from older roost configs (per scoping discussion).

## Output

On success:

```
Wrote .orchestrator/config.json
Wrote .orchestrator/.gitignore

Next step:
  roost spawn myproject-lead-pm \
    --channels '#myproject-leads' \
    --prompt '/lead-pm myproject <milestone> <your-irc-nick> <your-gh-login>'
```

The next-step line uses the resolved project name + channel + lead-pm slash-command shape so a haiku-following user can copy-paste.

## CLI surface

```
roost init                                            # autodetect everything
roost init --project P --repo O/R --agent-login L     # explicit overrides
roost init --agent-login user1 --agent-login user2    # multiple agent logins
roost init --force                                    # overwrite existing config
roost init --dry-run                                  # print what would be written, exit 0
```

## Implementation

Bash subcommand inside `bin/roost`, alongside the existing `spawn` / `shutdown` / `attach` / `tail` / `status` / `root` / `send` cases. Roughly:

- Inline arg parsing with the same `case` loop pattern as `spawn`. Repeatable `--agent-login` accumulates into a bash array.
- Preconditions: `git rev-parse --is-inside-work-tree`, `gh auth status`, `gh repo view`. All exit non-zero with a hint on failure.
- Autodetection helpers: `basename "$PWD" | tr '[:upper:]' '[:lower:]'`, `git remote get-url origin | sed -E 's|.*[:/]([^/:]+/[^/]+?)(\.git)?$|\1|'`, `gh api user --jq .login`.
- Project regex validation: `[[ "$project" =~ ^[a-z0-9][a-z0-9-]*$ ]]` — mirrors the regex in `src/orchestrator/naming.ts`. Cheap to keep in sync since it's six characters.
- JSON write: a single `printf` heredoc, same pattern as the settings.json generator at bin/roost:328. Our config is flat and the only string value carrying user input is `repo`, which we already validated via `gh repo view` — no need for jq.
- `.orchestrator/.gitignore`: plain `cat <<EOF`.
- `--dry-run`: print would-be contents to stdout, exit 0 without writing.
- Tests: a `test/init_test.sh` script that creates a temp git repo, runs `roost init` in non-interactive permutations, and asserts the resulting config.json + .gitignore. Hooks into `bun test` is not necessary — a bats-style script invoked from CI is fine.

Estimated size: ~80 lines of shell.

## Out of scope (followup issues if we want them)

- Watching initial issues: the user does that via DM to the watcher after `roost spawn`. Init does not seed `watched_issues`.
- Bringing up ergo / dispatcher: separate concern (#195's lazy-start design).
- Generating agent prompts / skill bindings: those ship in the plugin, not per-project.
- A `roost doctor` health check: separate followup.

## Why beta

The four-command get-started flow does not work without this. `roost init` is the bridge between `brew install` and `roost spawn`.

TeakBuilds:
Closes #202.

Adds `roost init` as a bash subcommand of `bin/roost`. No new deps.

**What it writes:**
- `.orchestrator/config.json` — project/repo/agent_logins/watched_issues/watched_prs (no irc fields; defaults in naming.ts)
- `.orchestrator/.gitignore` — state.json, last-tick.txt, last-error.txt
- `.claude/commands/{lead-pm,worker,reviewer,watcher}.md` — project-local prompt copies for customization

**Autodetect:** project from cwd basename, repo from git origin (`.git`-stripped, then `sed`-parsed), agent_login from `gh api user`. All overridable via flag.

**Preconditions:** git repo, gh auth, repo reachable via gh, project name matches `^[a-z0-9][a-z0-9-]*$`. All exit non-zero with actionable hints.

**Flags:**
- `--force` — overwrite config.json + .gitignore
- `--force-prompts` — overwrite existing prompt copies (independent of `--force`)
- `--no-prompts` — skip prompt copy entirely
- `--dry-run` — preview without writing (bypasses existing-config guard)

**Prompt copy:** idempotent by default (existing files skipped). Success output lists copied files + tip about customization and `--force-prompts`. No in-file banner — tip goes to stdout only.

**Tests:** `test/init_test.sh` — 20 cases: all four remote URL shapes, 0/1/2+ agent_logins, flag overrides, --force, --dry-run (including on existing config), prompt fresh-copy, skip-existing, --force-prompts overwrite, --no-prompts. Uses real git; stubs `gh` via PATH. shellcheck clean.

[roost-worker-202]

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -53,10 +53,10 @@ Options:
   --repo OWNER/REPO      GitHub repo (default: parsed from git origin remote).
   --agent-login LOGIN    GitHub login for agent_logins. Repeatable.
                          Default: current gh user.
-  --force                Overwrite existing .orchestrator/config.json and
-                         .orchestrator/.gitignore.
-  --force-prompts        Overwrite existing .claude/commands/ prompt copies.
-                         Independent of --force; does not touch config.json.
+  --force                Overwrite everything: config.json, .gitignore, and
+                         .claude/commands/ prompt copies.
+  --force-prompts        Overwrite only .claude/commands/ prompt copies.
+                         Does not touch config.json or .gitignore.
   --no-prompts           Skip copying prompts to .claude/commands/.
   --dry-run              Print what would be written; do not write files.
 
@@ -348,7 +348,7 @@ cmd_init() {
     for _pf in "${prompt_files[@]}"; do
       local _dst
       _dst="${commands_dir}/$(basename "$_pf" .md).md"
-      if [ ! -f "$_dst" ] || [ "$force_prompts" -eq 1 ]; then
+      if [ ! -f "$_dst" ] || [ "$force_prompts" -eq 1 ] || [ "$force" -eq 1 ]; then
         cp "$_pf" "$_dst"
         echo "Wrote ${_dst}"
         wrote_prompts=1
--- a/bin/roost
+++ b/bin/roost
@@ -355,9 +355,9 @@ cmd_init() {
       fi
     done
     if [ "$wrote_prompts" -eq 1 ]; then
-      printf '\nTip: prompts under %s/ are project-local copies. Edit them\n' "$commands_dir"
-      printf '     to fit your conventions, or delete a file to fall back to the\n'
-      printf "     plugin default. Re-run \`roost init --force-prompts\` to refresh.\n"
+      printf '\nTip: prompts under %s/ are project-local copies of the roost\n' "$commands_dir"
+      printf '     defaults. Edit them to fit your conventions. Re-run\n'
+      printf "     \`roost init --force-prompts\` to refresh from the plugin tree.\n"
     fi
   fi
 
--- a/bin/roost
+++ b/bin/roost
@@ -52,7 +52,11 @@ Options:
   --repo OWNER/REPO      GitHub repo (default: parsed from git origin remote).
   --agent-login LOGIN    GitHub login for agent_logins. Repeatable.
                          Default: current gh user.
-  --force                Overwrite existing .orchestrator/config.json.
+  --force                Overwrite existing .orchestrator/config.json and
+                         .orchestrator/.gitignore.
+  --force-prompts        Overwrite existing .claude/commands/ prompt copies.
+                         Independent of --force; does not touch config.json.
+  --no-prompts           Skip copying prompts to .claude/commands/.
   --dry-run              Print what would be written; do not write files.
 
 Examples:
@@ -60,6 +64,8 @@ Examples:
   roost init --project myapp --repo Owner/myapp
   roost init --agent-login alice --agent-login bob
   roost init --force
+  roost init --force-prompts
+  roost init --no-prompts
 EOF
 }
 
@@ -205,16 +211,20 @@ cmd_init() {
   local repo=""
   local agent_logins=()
   local force=0
+  local force_prompts=0
+  local no_prompts=0
   local dry_run=0
 
   while [ "$#" -gt 0 ]; do
     case "$1" in
-      --project)     project="$2";              shift 2 ;;
-      --repo)        repo="$2";                 shift 2 ;;
-      --agent-login) agent_logins+=("$2");      shift 2 ;;
-      --force)       force=1;                   shift ;;
-      --dry-run)     dry_run=1;                 shift ;;
-      -h|--help)     usage_init;                exit 0 ;;
+      --project)       project="$2";         shift 2 ;;
+      --repo)          repo="$2";            shift 2 ;;
+      --agent-login)   agent_logins+=("$2"); shift 2 ;;
+      --force)         force=1;              shift ;;
+      --force-prompts) force_prompts=1;      shift ;;
+      --no-prompts)    no_prompts=1;         shift ;;
+      --dry-run)       dry_run=1;            shift ;;
+      -h|--help)       usage_init;           exit 0 ;;
       *) echo "error: unknown init option: $1" >&2; exit 1 ;;
     esac
   done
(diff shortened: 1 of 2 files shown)


bin/roost
```diff
@@ -25,6 +26,7 @@ usage() {
 Usage: roost <command> [args]
 
 Commands:
+  init                     Bootstrap .orchestrator/config.json for a project.
```
AlexSc: Needs doc update

bin/roost
```diff
@@ -38,6 +40,35 @@ Run 'roost <command> --help' for per-command usage.
 EOF
 }
 
+usage_init() {
+  cat <<EOF
+Usage: roost init [options]
+
+Bootstrap .orchestrator/config.json for a new project. Autodetects project
```
AlexSc: Needs doc update

bin/roost
```diff
@@ -175,6 +206,161 @@ Example:
...
+    logins_json="${logins_json}${sep}\"${l}\""
+    sep=','
+  done
+  logins_json="${logins_json}]"
+
+  local config_path=".orchestrator/config.json"
+  local gitignore_path=".orchestrator/.gitignore"
+  local prompts_dir="${ROOST_DIR}/prompts"
+  local commands_dir=".claude/commands"
+  local prompt_names=(lead-pm worker reviewer watcher)
```
AlexSc: Would prefer a listing instead of duplicative hardcoding

AlexSc (changes requested):

bin/roost
```diff
@@ -175,6 +207,166 @@ Example:
...
+      _dst="${commands_dir}/$(basename "$_pf" .md).md"
+      if [ ! -f "$_dst" ] || [ "$force_prompts" -eq 1 ]; then
+        cp "$_pf" "$_dst"
+        echo "Wrote ${_dst}"
+        wrote_prompts=1
+      fi
+    done
+    if [ "$wrote_prompts" -eq 1 ]; then
+      printf '\nTip: prompts under %s/ are project-local copies. Edit them\n' "$commands_dir"
+      printf '     to fit your conventions, or delete a file to fall back to the\n'
```
AlexSc: We're not surfacing prompts in our plugin, unless I missed something, Nor was I intending to

bin/roost
```diff
@@ -38,6 +40,36 @@ Run 'roost <command> --help' for per-command usage.
...
+Bootstrap .orchestrator/config.json + .gitignore for a new project and copy
+roost prompt templates to .claude/commands/ for local customization.
+Autodetects project name, repo, and GitHub user; each value can be overridden.
+
+Options:
+  --project NAME         Project name (default: basename of cwd, lowercased).
+  --repo OWNER/REPO      GitHub repo (default: parsed from git origin remote).
+  --agent-login LOGIN    GitHub login for agent_logins. Repeatable.
+                         Default: current gh user.
+  --force                Overwrite existing .orchestrator/config.json and
```
AlexSc: Does --force imply --force-prompts? Noting a lot of orchestrator/prompts not being aligned

AlexSc (approved):

PR #217, opened 2026-05-10
feat: proxy AskUserQuestion over IRC via permbot (#208)

### LINKED ISSUE #208: Hook: proxy AskUserQuestion over IRC
## Problem
When an agent calls the `AskUserQuestion` tool, the question + options render in the Claude Code UI. For roost agents that's invisible to humans on IRC — the agent appears to be hung waiting for a UI response that nobody on IRC will give.

Hit live in the rebrand thread: lead-pm asked about repo visibility via AskUserQuestion, blocked silently until alex noticed.

## Proposed fix
A `PreToolUse` hook with matcher `AskUserQuestion` that:
1. Receives the question + options as JSON via stdin (Claude Code passes tool input to the hook)
2. Posts a multiline IRC message to the agent's project channel ("question + each option as a numbered choice")
3. Waits for a reply matching the channel/sender (similar to how `bin/irc-permission-prompt` waits for permbot replies)
4. Maps the reply back to the chosen option and returns it as the "user response" via hook output

## Open design questions
- Reply syntax: `1`/`2`/option-label/free-text? Probably accept any of those.
- Timeout behavior: if no IRC reply in N minutes, fall back to UI? Just block forever? Block + notify watcher?
- Multi-select questions: comma-separated reply (`1,3`)? Multiple messages?
- Where does the hook talk to IRC? Likely via the existing roost-irc MCP socket or an `roost send`-style helper.

## Constraints
- Must be opt-in via `bin/roost` settings.json wiring (not a global Claude Code hook)
- Must handle the case where IRC is disconnected gracefully (fall through to UI? error out?)
- Should be fast on the happy path (<1s overhead) since it intercepts every AskUserQuestion call

## Out of scope
- Replacing the full Claude Code permission flow (that's `bin/irc-permission-prompt`'s job, separate)
- Any changes to the Claude Code core

TeakBuilds:
closes #208

## Summary

- `bin/irc-ask-question` + `src/ask-question-hook.ts`: PreToolUse hook that intercepts AskUserQuestion, formats question+options as a numbered list, sends to permbot socket with `channel` + `replyTarget`, waits for reply, maps it to an option label, and returns `permissionDecision:allow` + `updatedInput.answers` — skipping the terminal UI entirely
- `src/permbot.ts`: extended with `channel` (post to channel vs DM target) and `replyTarget` (per-request reply nick) socket fields; accepts in-channel replies from target in addition to DMs; auto-joins ask channel on first request
- `src/irc-server.ts`: starts permbot when `ROOST_ASK_TARGET` is set without `ROOST_PERM_TARGET`
- `bin/roost`: `--ask-irc CHANNEL` and `--ask-target NICK`; shares existing permbot socket with `--perm-irc`
- `test/ask-question-hook.test.ts` + `test/permbot.test.ts`: unit tests for reply parsing, question formatting, channel routing, in-channel/DM reply acceptance, auto-join

On timeout: deny with error message (not silent UI fallthrough).

### DIFF
--- a/src/permbot.ts
+++ b/src/permbot.ts
@@ -13,7 +13,6 @@ import type { IrcMessage, RoostIrcClient } from './irc-client.js'
 export interface PermbotConfig {
   nick: string
   sockPath: string
-  target: string
   worker: string
   // Defaults to <dirname(sockPath)>/permbot.log when omitted. Pass '/dev/null'
   // in tests to silence file output.
@@ -34,10 +33,8 @@ interface QueueEntry {
   summary: string
   timeout: number
   kind: 'permission' | 'question'
-  // For kind='question': channel to post to; accept in-channel replies in addition to DMs.
-  channel?: string
-  // When set: accept replies from this nick instead of config.target.
-  replyTarget?: string
+  replyTarget: string  // required on every request
+  channel?: string     // set → channel mode (post + accept in-channel); absent → DM mode
 }
 
 interface InFlight {
@@ -79,7 +76,7 @@ export function startPermbot(
   config: PermbotConfig,
   client: RoostIrcClient,
 ): { stop: () => void; ready: Promise<void> } {
-  const { nick, sockPath, target, worker } = config
+  const { nick, sockPath, worker } = config
   const debugLog = config.debugLog ?? path.join(path.dirname(sockPath), 'permbot.log')
   const nudgeAfterMs = config.nudgeAfterMs ?? 5 * 60 * 1000
   const log = (msg: string) => dlog(debugLog, nick, msg)
@@ -96,9 +93,9 @@ export function startPermbot(
   function maybeDispatch(): void {
     if (inFlight !== null || queue.length === 0 || shuttingDown) return
     const entry = queue.shift()!
-    const { kind } = entry
-    const postTarget = entry.channel ?? target
-    const effectiveReplyTarget = (entry.replyTarget ?? target).toLowerCase()
+    const { kind, replyTarget } = entry
+    const postTarget = entry.channel ?? replyTarget
+    const replyTargetLc = replyTarget.toLowerCase()
     const summaryLines = entry.summary.split('\n').map(l => l.trimEnd()).filter(l => l.trim())
 
     const lines = [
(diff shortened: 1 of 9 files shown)


AlexSc (changes requested):
I feel like default target with override option is a bit of a path to confusion and madness. Thoughts on always requiring an explicit target nick and dm or channel for permbot internals?

src/permbot.ts
```diff
@@ -13,24 +13,51 @@ import type { IrcMessage, RoostIrcClient } from './irc-client.js'
...
   // Defaults to <dirname(sockPath)>/permbot.log when omitted. Pass '/dev/null'
   // in tests to silence file output.
   debugLog?: string
   nudgeAfterMs?: number  // default: 5 minutes; exposed for testing
 }
 
+interface PermBotRequest {
+  summary?: unknown
+  timeout?: unknown
+  kind?: unknown   // 'permission' | 'question'; falls back to channel-presence if absent
```
AlexSc: Inclined to make required instead of silent fallback

src/permbot.ts
```diff
@@ -13,24 +13,51 @@ import type { IrcMessage, RoostIrcClient } from './irc-client.js'
...
   sockPath: string
-  target: string
   worker: string
   // Defaults to <dirname(sockPath)>/permbot.log when omitted. Pass '/dev/null'
   // in tests to silence file output.
   debugLog?: string
   nudgeAfterMs?: number  // default: 5 minutes; exposed for testing
 }
 
+interface PermBotRequest {
```
AlexSc, line 24: Can any of these actually be undefined or unknown?

src/permbot.ts
```diff
@@ -128,18 +164,37 @@ export function startPermbot(
...
-      queue.push({ socket, summary, timeout })
+      const channel = typeof req.channel === 'string' && req.channel ? req.channel.toLowerCase() : undefined
+      const replyTarget = typeof req.replyTarget === 'string' && req.replyTarget ? req.replyTarget : null
+      if (!replyTarget) {
+        log('request missing replyTarget — rejected')
+        respond(socket, { error: 'replyTarget is required' })
+        return
+      }
+      // Explicit kind takes precedence; fall back to channel-presence for backward compat.
+      const kind: 'permission' | 'question' = req.kind === 'question' || req.kind === 'permission'
```
AlexSc: let's make a proper union type instead of inlining the type here, reads ugly

AlexSc (changes requested):
Can we also get an opus reviewer on this to look for opportunities to simplify all of permbot? This is a big responsibility expansion and some rearchitecting, so anything is on the table to make the code easier to follow and lean more towards a small API surface

AlexSc (approved):

Issue #213, opened 2026-05-10
README: surface brew install as recommended path


TeakBuilds:
The README's Setup section walks through manual prerequisites (bun, ergo download) + the `/plugin marketplace add` flow. With `avesalight/tap/roost` shipped (brings bun, tmux, ergo as deps and installs the plugin tree), the brew path should be the documented recommended install.

## Scope

- Add a "Install via Homebrew" section as the primary install path: `brew install avesalight/tap/roost`.
- Note the `brew tap oven-sh/bun` prerequisite (auto-tap of qualified deps is unreliable).
- Keep the marketplace install flow as an alternative for plugin-only / source installs.
- Trim Prerequisites: under brew you don't manually install bun/tmux/ergo.
- Drop the manual ergo download step in favor of `ergo run --conf "$(roost root)/etc/ergo.yaml"` once roost is brew-installed (ergo is on PATH).

## Out of scope

- Reorganizing the rest of the README (architecture, MCP tool reference, etc.). Only the Prerequisites + Setup sections.
- Changes to `roost init` behavior.

## Acceptance

- README's first install path is brew. A reader following it from scratch can `brew install avesalight/tap/roost`, start ergo via `ergo run --conf ...`, run `roost spawn`, and have a working session.
- Marketplace install remains documented but de-emphasized.

PR #221, opened 2026-05-10
docs: add 'previewing assets' tip to CLAUDE.md

TeakBuilds:
## Summary

- Adds a `## Previewing assets` section to roost's CLAUDE.md
- Documents the macOS / Linux / cross-platform OS tools for viewing files when sharing assets over IRC (qlmanage / xdg-open / feh)
- Mirrors the matching section just added to the brand repo's CLAUDE.md

## Why

From the #213 dogfooding postmortem: visual-context-over-text is real strain in IRC, and we briefly considered building a `channel_image` MCP / static-server bridge before alex pointed out IRC clients in tmux don't inline-preview anything. The right answer is to use the operator's existing OS tools, not bolt new abstractions onto IRC. The round-trip cost is congruent with the brand.

## Test plan

- [ ] CLAUDE.md renders cleanly on GitHub
- [ ] No other content shifted


AlexSc:
Honestly I don't see any reason to add this. It's not like you can show me images through your TUI anyways so this isn't a new limitation

PR #225, opened 2026-05-10
prompts/lead-pm: default reviewer to opus

TeakBuilds:
## Summary
- Step 6: opus is now the default for reviewer spawn, not a "if >250 lines" upgrade
- Sonnet kept as an escape hatch for trivially-sized doc/prompt PRs (<100 lines)

## Why
#208 postmortem: opus surfaced 14 findings on PR #217 that the original sonnet review didn't catch — duplicated `CHAT_KEYWORDS` with "must stay in sync" comments, the misleading "backward compat" branch that was actually the live path, and the dead dynamic-channel auto-join. Cost of opus review is small relative to a stale comment shipping.

## Test plan
- [x] Prompt-only change, no code paths affected

### DIFF
--- a/prompts/lead-pm.md
+++ b/prompts/lead-pm.md
@@ -73,7 +73,7 @@ To work on an issue:
   - Does it believably resolve the issue?
   - Does it set the project up for downstream success, or is it a pending footgun?
   - When worker proposes "X is fine for now" and you can already see a real gap, push back before approving the plan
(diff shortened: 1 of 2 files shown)


AlexSc:
Can you do some web searching to see if there are high quality reviewer prompts we could incorporate? The built in's are very diff focused, I feel like something that pushes to consider how the work slots into the whole would help

AlexSc (approved):

PR #229, opened 2026-05-10
feat: add hint to unread postfix explaining how to clear

### LINKED ISSUE #216: Add reminder to unread postfix that responding or acking will clear unread
I'm noticing a lead agent getting repeated unread reminders without addressing them, so we should include a quick inline "here's what to do about this"

TeakBuilds:
Closes #216

Agents were getting repeated unread reminders without any hint about how to act on them. Both the inline `unread:` postfix (appended to channel_message / channel_ack / direct_message responses) and the `unread-summary` push notification now end with:

> (post a message to that channel/peer or call channel_ack to clear)

Covers channel and DM peers. No new tests needed — existing `toContain` assertions all still pass (269/269).


AlexSc:
As long as we're here, do we clear unread for a channel on leaving it?

AlexSc (approved):
Lead, 220 next please

PR #242, opened 2026-05-11
feat: warn once when watched PR has empty linked_issues

### LINKED ISSUE #227: dispatcher: warn once when a watched PR has empty linked_issues
Caught from scratcher's dogfooding session. PR #190 had no closing keyword in its body, so GitHub reported empty \`linked_issues\`. Dispatcher had no \`#project-issue-N\` channel to route the PR's events to and silently dropped them — including a CHANGES_REQUESTED review. The scratcher lead-pm only found out by gh-fetching manually.

The fix isn't about pulling the human in — it's about giving the lead-pm (in the project channel) a signal it can act on. Lead-pms can edit PR bodies; once the closing keyword is added, GitHub auto-links and routing resumes.

## Fix

When the dispatcher first sees a watched PR with \`linked_issues: []\`, emit a one-time warning to the project channel:

> ⚠️ PR #N has no linked issues — events from it won't be routed. Add \`Closes #M\` (or Fixes/Resolves) to the PR body so the dispatcher can route.

Audience is the lead-pm sitting in the project channel; it's an agent-to-agent signal for self-recovery, not a human-attention ceremony.

\"One-time\" via a flag on the in-state PR snapshot (\`warned_no_linked\` or similar). Self-healing: once linked_issues becomes non-empty on a subsequent poll, clear the flag so a re-occurrence would warn again. Don't silently fall back to a default channel — that hides the root cause forever and the lead-pm has no signal to act on.

Companion: #226 closes the loophole on the prompt side by nudging workers to use a closing keyword in the first place.

## Acceptance
- New event kind or warning relay path for empty-linked-issues
- Posts to project channel exactly once per PR per state-snapshot of \"empty\"
- Clears when linked_issues becomes non-empty (so subsequent reset would re-warn)
- Test coverage in \`src/orchestrator/plugins/github/diff.test.ts\` or equivalent

TeakBuilds:
Closes #227

When a watched PR has no closing keyword, GitHub returns empty `linked_issues` and the dispatcher silently drops events (including CHANGES_REQUESTED reviews). This gives lead-pm a signal to act on.

## What changed

- `PrSnap` gets `warned_no_linked?: boolean` — set after warning, cleared when `linked_issues` becomes non-empty, absent=false (warns on first tick after upgrade)
- New `pr_no_linked_issues` event kind routed to the project channel (e.g. `#proj-leads`) instead of a PR-number-as-issue-channel fallback
- Warning emitted in `diffPr` (normal ticks) and `computePrEvents` (new PR added to watch)
- Format: `⚠️ PR org/repo#N has no linked issues — events won't be routed. Add Closes #N (or Fixes/Resolves) to the PR body: <url>`

## Tests

40 tests pass. New coverage: `diffPr — pr_no_linked_issues` (5 cases including upgrade/reset semantics), `computePrEvents — pr_no_linked_issues` (3 cases), `shouldPush`, and plugin routing to project channel.

### DIFF
--- a/src/orchestrator/plugins/github/format.ts
+++ b/src/orchestrator/plugins/github/format.ts
@@ -105,6 +105,10 @@ export function formatEvent(event: OrchestratorEvent): string {
     return `Issue ${tag} BACKLOG: ${ev.comment_count ?? 0} comments existed before watch — scan manually: ${event.url ?? ''}`
   }
 
+  if (kind === 'pr_no_linked_issues') {
+    return `⚠️ PR ${tag} has no linked issues — events won't be routed. Add Closes #N (or Fixes/Resolves) to the PR body: ${event.url ?? ''}`
+  }
+
   if (kind === 'dispatcher_error') {
     const ev = event as SeedEvent
     const tb = (ev.stderr ?? '').trim().split('\n')
--- a/src/orchestrator/plugins/github/format.ts
+++ b/src/orchestrator/plugins/github/format.ts
@@ -106,7 +106,7 @@ export function formatEvent(event: OrchestratorEvent): string {
   }
 
   if (kind === 'pr_no_linked_issues') {
-    return `⚠️ PR ${tag} has no linked issues — events won't be routed. Add Closes #<issue> (or Fixes/Resolves) to the PR body: ${event.url ?? ''}`
+    return `WARN PR ${tag} has no linked issues — events won't be routed. Add Closes #<issue> (or Fixes/Resolves) to the PR body: ${event.url ?? ''}`
   }
 
   if (kind === 'dispatcher_error') {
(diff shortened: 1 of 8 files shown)


AlexSc:
Quick q: will this still warn if there are explicit channels given?

Wondering if we should also add a quick p"now watching pr... routing to..." on successfully watching a PR as well

src/orchestrator/plugins/github/format.ts
```diff
@@ -105,6 +105,10 @@ export function formatEvent(event: OrchestratorEvent): string {
     return `Issue ${tag} BACKLOG: ${ev.comment_count ?? 0} comments existed before watch — scan manually: ${event.url ?? ''}`
   }
 
+  if (kind === 'pr_no_linked_issues') {
+    return `⚠️ PR ${tag} has no linked issues — events won't be routed. Add Closes #<issue> (or Fixes/Resolves) to the PR body: ${event.url ?? ''}`
```
AlexSc: No emoji in IRC

AlexSc (approved):
Nit but I don't need to rereview after fix. Lead please slot new issue into dag

PR #254, opened 2026-05-11
fix(test): wait for second multiline before history fetch (#252)

### LINKED ISSUE #252: Flaky test: channel_history multiline race — 'line two' arrives after history fetch
Caught during #250 CI debug — worker-235 reproduced 1/3 on main locally.

\`channel_history handles multiline\` test (\`test/irc-server.test.ts:236\`) asserts that the returned history contains \"line two\", but the test calls \`channel_history\` without waiting for the full multiline batch to be received by the MCP. If the second line arrives after the history snapshot, the assertion fails.

Belongs to the same family as #246 — schema-shape change (#239's XML history) layered onto a test that already had a latent ordering bug. The XML wrapping made the failure obvious; the race was there before.

## Fix sketch

Wait for both lines to land before calling channel_history. Either:
- await \`waitForNotification\` for the second message before the history call
- or use a sentinel non-multiline message after the multiline and wait for it

Either pattern matches how other multiline tests in this file already settle their state.

## Why not blocker for 1.0

Flake rate ~1/3 locally, retry-passes cleanly in CI. Real fix is small.

TeakBuilds:
Closes #252

## Summary
Two test races, both surfaced by coverage instrumentation timing.

**Race 1 — channel_history multiline shape (irc-server.test.ts)**: test waited for `line one` arrival before calling `channel_history`, then asserted on the full multiline body. Under coverage, `line two` sometimes hadn't been recorded yet. Fixed by waiting for `line two` — both lines are in the same batch, so once the second arrives we know the full multiline is in the buffer.

**Race 2 — long multiline reassemble (outbound.test.ts)**: the test used `Promise.all` to dispatch sender and receiver joins concurrently, then awaited sender seeing receiver's JOIN event. If receiver won the race, sender arrived into an already-populated channel and never saw the JOIN — predicate unsatisfiable, 15s timeout. Coverage shifted timing enough to make receiver win reliably. Sequence the joins.

## Test plan
- [x] Targeted tests pass 5/5 in isolation under coverage
- [x] Full suite passes 299/0 under coverage
- [ ] CI green

### DIFF
--- a/test/irc-server.test.ts
+++ b/test/irc-server.test.ts
@@ -228,7 +228,7 @@ describe.if(isErgoAvailable())('irc-server MCP tools', () => {
     await peer.joinChannel('#ip-histshape5')
 
     peer.say('#ip-histshape5', 'line one\nline two')
-    await mcp.waitForNotification(n => n.meta.channel === '#ip-histshape5' && n.content.includes('line one'))
+    await mcp.waitForNotification(n => n.meta.channel === '#ip-histshape5' && n.content.includes('line two'))
 
     const hist = await mcp.client.callTool({ name: 'channel_history', arguments: { channel: '#ip-histshape5' } })
     const text = toolText(hist)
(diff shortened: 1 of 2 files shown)


AlexSc:
If you see a test failure fix it ASAP. Don't worry about "scope". Unreliable tooling is a drop everything deal with it right now emergency

AlexSc (approved):

PR #264, opened 2026-05-11
maint: test suite fragility — typed wire meta + helpers (#246)

### LINKED ISSUE #246: test suite fragility: implicit wire-shape invariants + opaque CI hangs
Captured fresh from the #244 review cycle. Two compounding problems made a single attr change (\`event=\"message\"\` addition) cascade into a 24-min CI hang + cancel + multiple fixup commits before merge.

## Root cause: implicit wire-shape invariants

5 test sites in \`test/multiline-edge-cases.test.ts\` + 1 in \`test/outbound.test.ts\` filtered regular messages with \`!n.meta.event\`. One assertion in \`test/irc-server.test.ts:446\` was \`expect(msg.meta.event).toBeUndefined()\`. All encoded an invariant — *\"regular messages have no \`event\` attr\"* — that wasn't documented anywhere as load-bearing. The PR added \`event=\"message\"\` to make message-type a positive discriminator; every consumer reading absence-of-attr silently broke.

Marketing's read (which I'm stealing): the doc gap is *invariants the test suite quietly assumes*. Same trap is loaded for #237 (mention=true addition) and #239 (channel_history shape change).

## Compounding: bun + coverage makes failure invisible

When the predicate broke, tests hit the \`waitForNotification\` default 5s timeout. Under \`bun test --coverage\`, the runner stops emitting test events after a post-timeout assertion failure (locally exits 0 but truncates results; CI hangs the runner). The \"failure\" surfaces as a 24-min hang with no log output, not a clean fail.

This is partly upstream bun behavior, partly our 5s default being too tight for coverage-instrumented multiline tests.

## What to do (size-ordered)

**Cheap:**
- [ ] Document the meta shape contract in one place (probably \`src/irc-server.ts\` or \`docs/LEARNINGS.md\`) — list every \`meta.*\` attr and whether absence has semantic meaning
- [ ] Add a workflow timeout to \`.github/workflows/ci.yml\` so a hung test surfaces as failure at e.g. 8min rather than 22+min wedge
- [ ] CI-aware \`waitForNotification\` default (e.g. 15s under \`CI=1\`, 5s locally)

**Medium:**
- [ ] Typed discriminated union for meta — adding a new \`event\` value becomes a compile error at every consumer, not a runtime miss. Probably a \`ChannelMeta\` type covering the union of \`{event: 'message'}\`, \`{event: 'join'}\`, etc.
- [ ] Test helper like \`assertChannelMessage(n, {...})\` so the regular-message shape is asserted in one place — schema changes become a one-file edit

**File-upstream:**
- [ ] Bun bug report: \`--coverage\` + post-timeout assertion failure wedges event emission. See #170 for related deadlock. May or may not be the same root.

## Related

- #170 — original bun deadlock pattern + \`suppressLateRejection\` fix
- #172 — waitForMessage cleanup gap
- #245 — permbot socket promise wrap (rolled into #244; was a real bug but not the CI-hang root)

TeakBuilds:
Closes #246

## Summary

- **`src/wire-meta.ts`** — typed discriminated union for the `notifications/claude/channel` meta record. `pushNotification` now takes `WireMeta`; adding a new `event` value is a compile error at every emit site and every consumer that narrows on `event`. Wire bytes on the network are unchanged.
- **`.github/workflows/ci.yml`** — `timeout-minutes: 10` on the test job so a wedged run fails cleanly instead of running to GitHub's 6h default. The #246 trigger was a 24-min hang under `bun --coverage`.
- **`test/helpers/mcp-core.ts`** — `messagePredicate({...})` and `assertChannelMessage(n, {...})` share a single `matchesMessage` matcher. Migrated the 5 inline `event === 'message'` predicate sites in `multiline-edge-cases.test.ts` and `outbound.test.ts` to use the helper.
- **`docs/LEARNINGS.md`** — adds Finding H. The lesson (invariant-by-omission) is the durable bit; the contract itself lives in the TS types.

## Dropped from scope (per #246 checklist)

- CI-aware `waitForNotification` default — already addressed by 93b6a8d (5s → 15s blanket). Reverting to a CI/local split adds branchy code with no current pain.
- Bun upstream bug — research/file task, no code change here. Lead's call whether to split as a follow-up.

## Test plan

- [x] `bun run lint` clean
- [x] `bun test` — 304 pass, 0 fail
- [ ] CI green (will verify post-push)

### DIFF
--- a/src/wire-meta.ts
+++ b/src/wire-meta.ts
@@ -1,15 +1,22 @@
 // Wire shape of `notifications/claude/channel` meta records emitted by the MCP.
 //
-// The wire is `Record<string, string>` over JSON-RPC. These types are the
-// internal contract for the emit side: adding a new `event` value is a
-// compile-time error at every emit site, and the union forces every variant
-// to declare which attrs apply to it. Consumers (tests, the host harness)
-// can narrow by `event` for type-safe access to variant-specific attrs.
+// Emit-side contract: every call to pushNotification picks a variant of the
+// union, so adding a new `event` value is a compile error at every emit site.
+// The union also reuses `SystemKind` / `MembershipKind` from irc-client.ts —
+// the source of truth for those literals lives there, not duplicated here.
+//
+// Consumer-side caveat: the wire is `Record<string, string>` over JSON-RPC,
+// and `test/helpers/mcp-core.ts`'s `ChannelNotification.meta` is currently
+// typed loosely as `Record<string, string>`. Consumers can narrow against
+// `WireMeta` by importing it, but the helper layer does not enforce that.
+// Tightening it is a separate followup; see #246 review thread.
 //
 // Why string values: JSON-RPC serializes everything to strings on the wire,
 // so booleans render as the literal `'true'` and numbers as decimal strings.
 // Optional flags (e.g. `buffered?: 'true'`) are present-or-absent rather than
 // `'true' | 'false'` — keeps the wire compact and matches existing consumers.
+
+import type { SystemKind, MembershipKind } from './irc-client.js'
 
 export interface WireMessageMeta {
   event: 'message'
@@ -32,7 +39,7 @@ export interface WireReminderMeta {
 }
 
 export interface WireMembershipMeta {
-  event: 'join' | 'leave' | 'nick'
+  event: MembershipKind
   sender: string
   channel: string
   isDirect: 'false'
@@ -42,7 +49,7 @@ export interface WireMembershipMeta {
 }
 
 export interface WireSystemMeta {
-  event: 'disconnected' | 'reconnected' | 'cap-missing' | 'registered' | 'registration-failed'
+  event: SystemKind
   sender: string
   channel: string
   isDirect: 'false'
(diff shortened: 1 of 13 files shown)


src/wire-meta.ts
```diff
@@ -0,0 +1,76 @@
...
+// Emit-side contract: every call to pushNotification picks a variant of the
+// union, so adding a new `event` value is a compile error at every emit site.
+// The union also reuses `SystemKind` / `MembershipKind` from irc-client.ts —
+// the source of truth for those literals lives there, not duplicated here.
+//
+// Consumer-side caveat: the wire is `Record<string, string>` over JSON-RPC,
+// and `test/helpers/mcp-core.ts`'s `ChannelNotification.meta` is currently
+// typed loosely as `Record<string, string>`. Consumers can narrow against
+// `WireMeta` by importing it, but the helper layer does not enforce that.
+// Tightening it is a separate followup; see #246 review thread.
```
AlexSc: Comments should be timeless.

And if you have to explain "we'll do it later" maybe just do it instead

AlexSc (changes requested):
Comments should be timeless, and issues should be done or not done, not half done with deferred work

AlexSc (approved):

PR #269, opened 2026-05-11
feat: extract dispatcher boot+verify into bin/start-dispatcher

### LINKED ISSUE #260: Extract watcher dispatcher boot+verify into bin/start-dispatcher wrapper
## Followup from #258 / PR #259 review

Reviewer (opus, on PR #259) flagged:

> The root cause being patched is *haiku unreliability* (it picked the `roost` skill over the `roost` CLI). The fix moves the heavier verification logic *into* the prompt (7-line bash block, PID capture, kill-0, FAIL-string parse, conditional IRC post). Same fragile substrate, richer ask. A `bin/start-dispatcher` wrapper exiting 0/non-zero would push the brittle bit into deterministic shell and leave the prompt with one command + one branch. Not blocking; worth a follow-up.

#258 landed (PR #259) with the bash block inline in `prompts/watcher.md`. That's a reliability improvement over what was there, but the verification logic — PID capture, `sleep 2`, `kill -0`, FAIL parsing, conditional IRC notify — is in prompt text and depends on the model executing the block verbatim.

## What to do

Extract the boot-and-verify block into `bin/start-dispatcher`:
- Exits 0 if dispatcher came up cleanly, non-zero with a useful message otherwise
- Owns the stderr redirect to `dispatcher-boot.log`
- Owns the 2s sleep + `kill -0` check (and can grow to richer checks later — IRC join probe, socket ping, etc — without prompt edits)

Watcher prompt becomes a single line:

```
"$ROOST_DIR/bin/start-dispatcher" "$3" || (post failure to #leads; continue in degraded mode)
```

## Why this is post-1.0

#259 already makes the boot reliable enough for the milestone. This is hardening — it removes prompt brittleness as a class, not just the one instance we hit today. Right size for after 1.0.

## Related

- #258 (fixed by #259) — the underlying reliability issue
- prompts/watcher.md — current home of the bash block

TeakBuilds:
Closes #260

Extracts the inline bash block from `prompts/watcher.md` into `bin/start-dispatcher <config-dir>`.

- Exits 0 if dispatcher is alive after the 2s boot window
- Exits 1 with `start-dispatcher: FAIL: <reason>` on stderr if not
- Owns the `dispatcher-boot.log` redirect
- No idempotency guard (documented in script header) — watcher calls this once on boot

Watcher prompt is now one line instead of an inline bash block. The brittle verification logic lives in deterministic shell, not prompt text.

Homebrew formula bundles `bin/` wholesale via `libexec.install Dir["*", ".[!.]*"]` — no formula changes needed.

### DIFF
--- a/prompts/watcher.md
+++ b/prompts/watcher.md
@@ -46,7 +46,7 @@ A single message may contain multiple commands, separated by newlines, semicolon
 
 ## Behavior rules
 
-- **On boot, first action:** start and verify the dispatcher daemon (`--daemon` keeps it running; without it the orchestrator exits after one tick):
+- **On boot, first action:** start and verify the dispatcher daemon:
   ```bash
   "$ROOST_DIR/bin/start-dispatcher" "$3"
   ```
--- a/prompts/watcher.md
+++ b/prompts/watcher.md
@@ -46,15 +46,11 @@ A single message may contain multiple commands, separated by newlines, semicolon
 
 ## Behavior rules
 
-- **On boot, first action:** start and verify the dispatcher daemon (`--daemon` keeps it running; without it the orchestrator exits after one tick). Run this Bash block:
+- **On boot, first action:** start and verify the dispatcher daemon (`--daemon` keeps it running; without it the orchestrator exits after one tick):
   ```bash
-  nohup "$ROOST_DIR/bin/dispatcher" --daemon --config-dir "$3" >> "$3/dispatcher-boot.log" 2>&1 &
-  DISP_PID=$!
-  # 2s catches immediate failures (missing config, env, syntax); steady-state crashes are visible in daemon.log
-  sleep 2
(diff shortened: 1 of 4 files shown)


prompts/watcher.md
```diff
@@ -46,15 +46,11 @@ A single message may contain multiple commands, separated by newlines, semicolon
 
 ## Behavior rules
 
-- **On boot, first action:** start and verify the dispatcher daemon (`--daemon` keeps it running; without it the orchestrator exits after one tick). Run this Bash block:
+- **On boot, first action:** start and verify the dispatcher daemon (`--daemon` keeps it running; without it the orchestrator exits after one tick):
```
AlexSc: --daemon note now outdated unless the script takes it and I didn't notice

AlexSc (changes requested):

AlexSc:
Also is this new script covered by our shellcheck CI?

AlexSc (approved):

PR #270, opened 2026-05-11
prompts/worker: workflow primer + no ScheduleWakeup

### LINKED ISSUE #267: worker slash command: better workflow priming + suppress ScheduleWakeup usage
Two related gaps in `/worker` from this milestone's experience:

## Workflow priming

Workers don't know enough about the surrounding process they're a part of. Symptoms from #246/PR#264:
- Worker concluded "PR is draft per worker process — defer to you on flipping ready" after the lead-pm had already marked it ready. Worker had no internal model that the PR moves through draft→ready→review-loop and stays ready throughout.
- Worker's instinct on CI failure was "this must be a TS6/linux narrowing quirk" before checking whether the failure was actually in their code vs introduced upstream by a parallel merge.
- Worker treated lead-pm "categorize for in-PR vs followup" guidance as defer-by-default, when the lead's expectation was do-the-real-fix.

Prompt should convey: how leads pressure-test plans, that CI runs on the merge commit (so main can introduce failures), what "ready vs draft" means at each step, and that "followup" is the exception, not the default.

## ScheduleWakeup discouragement

Workers shouldn't be self-scheduling wakeups. They're driven by IRC notifications + lead direction; arbitrary timers add noise and waste budget. The prompt should explicitly tell workers not to call `ScheduleWakeup`.

Alex's note: he's tempted to remove the tool from worker by default — that's a separate, larger change (would require harness-level permission scoping). The prompt change is the cheap first step.

## Acceptance

- `skills/worker/SKILL.md` updated with both: workflow primer + explicit no-ScheduleWakeup line.
- Quick smoke: spawn a worker on a trivial issue, verify they don't call ScheduleWakeup and behave consistently around draft/ready.

TeakBuilds:
Closes #267

Adds four sections to `prompts/worker.md`:

- **PR lifecycle** — draft→ready→review loop, worker signals state, lead flips the bit, PR stays ready once marked
- **CI failures** — triage order: upstream drift → env mismatch → real bug
- **Plans and followups** — lead will pressure-test; in-PR is the default, followup only when lead names it or scope is genuinely orthogonal
- **Scheduling** — hard rule: do not call `ScheduleWakeup`

All four gaps surfaced from the #246/PR#264 experience.

### DIFF
--- a/prompts/worker.md
+++ b/prompts/worker.md
@@ -19,7 +19,7 @@ Ask in the channel before any destructive or shared-state action: force-push, br
 
 ## PR lifecycle
 
-PRs start as draft. When your work is complete and CI is green, signal clearly in the channel ("pushed, ready for review" or "addressed X, ready to flip"). Lead-pm then marks it ready. Once the PR is marked ready it stays ready through the review loop — you are done with draft/ready transitions. If a reviewer asks for changes, push the fix and say so; lead-pm will re-evaluate state.
+PRs start as draft. When your work is pushed, signal clearly in the channel ("pushed, ready for review" or "addressed X, ready to flip"). Lead-pm then marks it ready. Once the PR is marked ready it stays ready through the review loop — you are done with draft/ready transitions. If a reviewer asks for changes, push the fix and say so; lead-pm will re-evaluate state.
 
 ## CI failures
 
--- a/prompts/worker.md
+++ b/prompts/worker.md
@@ -8,7 +8,7 @@ Your task: GitHub issue $2#$1. Branch `$3` is checked out here.
 
 Process:
 1. Read the issue $2#$1 thoroughly — body, comments, labels, milestones, and any blocking relationships. `gh issue view $1 --comments` is the minimum (plain `gh issue view` skips comments, which often carry the actual scope). If your project provides a `github-management` skill, use it for richer output. Then read any relevant code.
-2. Post your implementation plan in #$0-issue-$1 and wait for lead-pm's approval before coding
+2. Post your implementation plan in #$0-issue-$1 and **wait** for lead-pm's approval before coding
 3. When done, open a *draft* PR and post the link in #$0-issue-$1. The PR body **must** start with a closing keyword on its own line — `Closes #$1` (or `Fixes` / `Resolves`). GitHub only auto-links issues when one of those keywords precedes the number; without it, `linked_issues` comes back empty and the dispatcher has no channel to route per-PR events to.
 4. Prefix all GitHub comments with [$0-worker-$1]
 5. Defer to lead-pm for marking the PR ready, tagging reviewers, and creating followup issues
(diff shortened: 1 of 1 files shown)


prompts/worker.md
```diff
@@ -16,3 +16,22 @@ Process:
 Don't mark the PR ready yourself.
 
 Ask in the channel before any destructive or shared-state action: force-push, branch deletion, hook bypass (`--no-verify`), `git reset --hard`, dropping unfamiliar files, or anything else that's hard to reverse. Local edits and pushes to your own feature branch don't need confirmation.
+
+## PR lifecycle
+
+PRs start as draft. When your work is complete and CI is green, signal clearly in the channel ("pushed, ready for review" or "addressed X, ready to flip"). Lead-pm then marks it ready. Once the PR is marked ready it stays ready through the review loop — you are done with draft/ready transitions. If a reviewer asks for changes, push the fix and say so; lead-pm will re-evaluate state.
```
AlexSc: Just "when work is pushed". The dispatcher will notify when CI is green, we're trying to discourage that self check behavior

prompts/worker.md
```diff
@@ -16,3 +16,22 @@ Process:
...
+
+## CI failures
```
AlexSc: Cut this section, may not apply to other projects

prompts/worker.md
```diff
@@ -16,3 +16,22 @@ Process:
...
+
+When CI fails, triage in this order:
+- **(a) Upstream drift** — did main advance since you branched? CI tests the merge commit, so a parallel merge can introduce failures unrelated to your change. Run `git fetch && git log HEAD..origin/main --oneline` to see what CI pulled in that you don't have.
+- **(b) Environment mismatch** — does the failure reproduce locally? CI may run a different OS, toolchain version, or stricter type flags.
+- **(c) Real bug** — only after ruling out (a) and (b), assume it's in your change.
+
+## Plans and followups
+
+Lead-pm will pressure-test your plan before approving. Have answers ready: why this approach, what alternatives were ruled out, what the edge cases are. If lead says "categorize for in-PR vs followup", the in-PR side is the default. File a followup only when the lead explicitly names one or when the scope is genuinely orthogonal (different file, different system, clearly separable concern).
```
AlexSc: PM should always be doing follow up filing. Worker should be told to prefer taking on more work.

prompts/worker.md
```diff
@@ -16,3 +16,22 @@ Process:
...
+
+## Scheduling
+
+Do not call `ScheduleWakeup`. You are driven by IRC notifications and lead direction. If you have nothing to do, sit idle — the lead will redirect you.
```
AlexSc: This reads a little harsh.

AlexSc (changes requested):

prompts/worker.md
AlexSc: Bold wait

AlexSc (approved):
One nit, plz address but I don't need to rereview

PR #277, opened 2026-05-12
feat(perm-irc): PreToolUse Bash hook to close safety-analyzer bypass

### LINKED ISSUE #276: PreToolUse hook to close the bash safety-check bypass in perm-irc
## Problem

Claude Code's `PermissionRequest` hook (which roost's `irc-permission-prompt` plugs into) does not fire for Bash tool calls that trip the harness's structural safety analyzer. When the bash analyzer flags a command — newline-hash, multi-cd, sed-dangerous, process-substitution, shell-operators, cd-git-compound, etc. — Claude Code shows a terminal-only prompt and skips `executePermissionRequestHooks` entirely. Workers spawned with `--perm-irc` hang silently on the local TUI instead of routing to the operator over IRC.

Observed today: `scratcher-worker-202` blocked ~27 minutes on a `bundle exec ruby -e \"…\n#…\"` heredoc that hit the `bashMissKind:\"semantics\"` / newline-hash safety check. Permbot log silent, transcript has no `hook_permission_decision` attachment for the call, hook process never spawned.

## Approach

Implement a `PreToolUse` hook with a `Bash` matcher. `PreToolUse` fires before the bash analyzer and can return `permissionDecision: \"allow\" | \"deny\" | \"ask\" | \"defer\"` via `hookSpecificOutput`. A `\"defer\"` decision falls through to the normal pipeline, so we only need to take a position when we know the analyzer is about to bypass `PermissionRequest`.

Two viable shapes:

- **Narrow:** detect just the safety-check patterns and route via IRC for those; defer otherwise.
- **Wide:** every Bash call goes through PreToolUse + IRC. Replaces PermissionRequest's role for Bash entirely; PermissionRequest stays for Write/Edit/etc.

Narrow is less code surface; wide is more uniform and dodges any future safety-check kinds Claude Code adds.

Reuses the existing permbot socket + IRC client. New file: `src/pretooluse-prompt.ts` (and `bin/irc-pretooluse-prompt`). Wire in `bin/roost` next to the existing `PermissionRequest` hook entry.

## Safety-check triggers

Pulled from the 2.1.139 binary. Useful as test fixtures: each is a one-line bash that lands in the safety-check path and skips `PermissionRequest`. (`bashMissKind` values come back on the `decisionReason`.)

| kind | trigger | example |
|---|---|---|
| `semantics` (newline-hash) | regex `/\n[ \t]*#/` inside quoted arg, env var value, or redirect target | `bash -c \"echo hi\n# comment\"` |
| `process-substitution` | `>(...)` or `<(...)` in command | `diff <(echo a) <(echo b)` |
| `multi-cd` | more than one `cd`/`pushd`/`popd` subcommand | `cd /tmp && ls && cd /var` |
| `cd-git-compound` | cd + git in one command, target dir ≠ cwd | `cd /tmp/repo && git status` |
| `cd-compound-write` | compound with cd + write op | `cd /tmp && echo x > out` |
| `cd-compound-redirect` | compound with cd + output redirection | `cd /tmp && ls > out` |
| `cd-multi-positional` | zsh two-arg `cd OLD NEW` | `cd OLD NEW` |
| `sed-dangerous` | sed with `w`, `e`, or similar dangerous ops | `sed 'w /tmp/out' input` |
| `shell-operators` | subshell `(...)` or command group `{...}` | `(ls; pwd)` |
| `flag-validation` | wrapper command with unvalidated flags | `env --chdir=/tmp ls` |
| `too-complex` | AST parser hit an unparseable construct | `echo $((x+1))` (arithmetic with non-literal) |

`no-rule-match` and `prompt-ask-rule` (the policy-level kinds) already go through `PermissionRequest` fine and aren't relevant here.

## Testing guidance

Two test layers, mirroring how `irc-permission-prompt` is tested today:

**1. Unit tests on the hook script** (`test/pretooluse-prompt.test.ts`)

Pattern matches `test/irc-permission-prompt.test.ts`: spin a fake IRC server, `Bun.spawn` the hook with a payload on stdin, assert the JSON on stdout. PreToolUse hook input shape:

```json
{
  \"session_id\": \"…\",
  \"transcript_path\": \"/abs/path.jsonl\",
  \"cwd\": \"/abs/cwd\",
  \"hook_event_name\": \"PreToolUse\",
  \"tool_name\": \"Bash\",
  \"tool_input\": { \"command\": \"…\", \"description\": \"…\" }
}
```

Expected hook output for an allow:

```json
{
  \"hookSpecificOutput\": {
    \"hookEventName\": \"PreToolUse\",
    \"permissionDecision\": \"allow\",
    \"permissionDecisionReason\": \"operator approved via IRC\"
  }
}
```

Decisions to cover: `defer` (non-Bash tool, or Bash that doesn't need routing), `allow` (operator y/yes), `deny` (operator n/no), `ask` (socket timeout fallback — same defer-to-terminal behavior as the existing hook).

**2. Integration smoke tests** (manual or scripted)

Spawn a worker with the new hook wired, fire each trigger from the table above, verify:
- transcript records a `hook_permission_decision` attachment for the toolUseID
- permbot.log shows a `client accepted` + queued request + reply
- worker proceeds (or aborts on deny) without ever showing a TUI prompt

Trigger expansion is the load-bearing piece. If we have a one-liner per `bashMissKind` in a fixture file, the smoke test is just a loop.

**3. Cross-version pin**

`bashMissKind` values were extracted from Claude Code 2.1.139. They're not part of the public hook contract, so the table could drift. The hook doesn't *depend* on the value being `\"semantics\"` (we route based on the tool name + command shape, not the harness's classification), but a `claude --version` floor in the smoke-test runner is worth adding so we notice if the safety-analyzer behavior changes.

## Out of scope

- Migrating the existing `PermissionRequest` hook's logic into PreToolUse wholesale (separate decision: do we keep both, or consolidate). Pick one when narrow vs. wide above is settled.
- Anything about Write/Edit/Read — those go through normal permission rules and `PermissionRequest` cleanly today.

Refs: bug surfaced in #roost-leads debug session 2026-05-11; related cleanup in #275.

TeakBuilds:
Closes #276

## Summary
- New `PreToolUse` hook with a `Bash` matcher (`src/pretooluse-prompt.ts`, `bin/irc-pretooluse-prompt`). Fires *before* the safety analyzer, so commands shaped like the `bashMissKind` triggers route through IRC instead of hanging silently on the terminal-only prompt.
- Narrow scope: `classifyBash(command)` is a pure function that returns the matching `BashMissKind` or null. Null defers — the allowlist + `PermissionRequest` pipeline continues untouched, so routine `ls` / `gh pr view` / `bun test` don't suddenly start pinging the operator.
- Reuses the permbot socket + DM fallback from `permission-prompt.ts`; no new IRC plumbing.
- Wired in `bin/roost` behind `--perm-irc`, merged into the same `PreToolUse` array as the existing `AskUserQuestion` matcher.

## Coupling note
`bashMissKind` values come from the 2.1.139 binary and aren't part of the public hook contract. The issue table is the spec; one test fixture per kind keeps drift visible (new kind → worker hang regression → add a pattern). Detection over-matches where the AST is ambiguous (e.g. any `cd` followed by `&&`/`||`/`;` triggers `cd-compound` rather than parsing for write/redirect/git specifically). False positives ping the operator unnecessarily; false negatives reproduce the original bug.

`BashMissKind` (the TypeScript type) uses roost's *approximations* of those labels — see the doc comment at the type. They're not literal `decisionReason` strings.

## Followup
- #278 — cross-file consistency between `permission-prompt.ts` and `pretooluse-prompt.ts` (env-var timeout override, default reasons, unrecognized-reply test); deferred so it doesn't drag the other hook into this diff.

## Test plan
- [x] `bun run lint` — clean
- [x] `bun run typecheck` — clean
- [x] `bun test test/pretooluse-prompt.test.ts` — 36/36 pass (classify coverage for each bashMissKind + subprocess allow/deny/ask/passthrough/owner-gate/unrecognized-reply/default-reason-on-allow)
- [x] `bun test` — 342/342 pass, no regressions
- [x] integration smoke: ran `claude -p` against this hook + a stub permbot socket auto-replying `y`. Both `cd /tmp && echo … > out` (cd-compound) and `diff <(echo a) <(echo b) > out` (process-substitution) routed through the hook and executed; a benign `ls /tmp` deferred to normal permission rules without ever hitting the stub. End-to-end proves the analyzer bypass is closed without regressing the allowlist.

### DIFF
--- a/src/pretooluse-prompt.ts
+++ b/src/pretooluse-prompt.ts
@@ -39,19 +39,20 @@ const SOCKET_SAFETY_TIMEOUT = Math.min(570, Math.max(1, Number(process.env['ROOS
 // `decisionReason` strings. We collapse several harness kinds (e.g.
 // cd-git-compound / cd-compound-write / cd-compound-redirect → `cd-compound`)
 // and rename others (`semantics` → `newline-hash` for legibility in operator
-// summaries). Don't grep the Claude Code binary for these — they won't be
-// there. Tests pin one example per kind from the issue table so any
-// classifier drift surfaces as a test failure.
+// summaries). The literal Claude Code strings are listed in the trailing
+// comment on each variant so a grep for the CC binary's `bashMissKind` value
+// (e.g. `cd-git-compound`) lands here. Tests pin one example per kind from
+// the issue table so any classifier drift surfaces as a test failure.
 export type BashMissKind =
-  | 'newline-hash'
-  | 'process-substitution'
-  | 'multi-cd'
-  | 'cd-compound'
-  | 'cd-multi-positional'
-  | 'sed-dangerous'
-  | 'shell-operators'
-  | 'flag-validation'
-  | 'too-complex'
+  | 'newline-hash'          // CC: "semantics"
+  | 'process-substitution'  // CC: "process-substitution"
+  | 'multi-cd'              // CC: "multi-cd"
+  | 'cd-compound'           // CC: "cd-git-compound" | "cd-compound-write" | "cd-compound-redirect"
+  | 'cd-multi-positional'   // CC: "cd-multi-positional"
+  | 'sed-dangerous'         // CC: "sed-dangerous"
+  | 'shell-operators'       // CC: "shell-operators"
+  | 'flag-validation'       // CC: "flag-validation"
+  | 'too-complex'           // CC: "too-complex"
 
 /**
  * Returns the bashMissKind a command resembles, or null if it should pass
@@ -62,10 +63,11 @@ export type BashMissKind =
 export function classifyBash(command: string): BashMissKind | null {
   if (!command) return null
 
-  // newline-hash: literal newline followed by optional whitespace then '#'.
-  // The original bug (worker-202 27-min hang) was a heredoc with this shape.
-  // Checked against the raw command — the analyzer's trigger explicitly
-  // requires the newline to be inside a quoted arg, env value, or redirect.
+  // CC bashMissKind: "semantics" (newline-hash).
+  // Literal newline followed by optional whitespace then '#'. The original
+  // bug (worker-202 27-min hang) was a heredoc with this shape. Checked
+  // against the raw command — the analyzer's trigger explicitly requires
+  // the newline to be inside a quoted arg, env value, or redirect.
   if (/\n[ \t]*#/.test(command)) return 'newline-hash'
 
   // Command-start anchor for cd patterns: start-of-string, shell operator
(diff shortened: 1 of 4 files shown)


AlexSc:
Have we done proper e2e testing to confirm this works and doesn't trigger false positives?

src/pretooluse-prompt.ts
```diff
@@ -0,0 +1,199 @@
...
+
+// bashMissKind labels lifted from the 2.1.139 binary (issue #276). These are
+// roost's *approximations* of the harness's classification — not literal
+// `decisionReason` strings. We collapse several harness kinds (e.g.
+// cd-git-compound / cd-compound-write / cd-compound-redirect → `cd-compound`)
+// and rename others (`semantics` → `newline-hash` for legibility in operator
+// summaries). Don't grep the Claude Code binary for these — they won't be
+// there. Tests pin one example per kind from the issue table so any
+// classifier drift surfaces as a test failure.
+export type BashMissKind =
```
AlexSc, line 46: Inline comment the Claude code strings for searching

AlexSc (approved):

PR #286, opened 2026-05-13
feat(config): writeConfig + mutateConfig with advisory lock

### LINKED ISSUE #284: dispatcher: writeConfig with atomic temp+rename and advisory lock
Today `.orchestrator/config.json` is operator-managed (or read+write by the watcher LLM). With #N+1 the dispatcher will mutate it from inbound DM handlers — concurrent writes need to not clobber each other or the poll loop's `loadConfig`.

## Scope

Add a `writeConfig` to `src/orchestrator/config.ts` that:
- writes to `.config.<pid>.<ts>.tmp` then atomically renames to `config.json` (mirror of `writeState`)
- holds an advisory lock (`config.lock` file) for the read-modify-write window so two DMs landing in the same tick don't race
- preserves field order via the existing `sortedJson` helper so diffs stay readable

Plus a `mutateConfig(stateDir, fn)` helper that takes the lock, loads, applies `fn(config)`, writes — single call site for all future mutators.

## Out of scope

- The DM command handler itself (#N+1).
- Migrating any existing operator-edited config.json (no migration needed; same schema).

## Acceptance

- `bun test` adds coverage for: concurrent `mutateConfig` calls serialize, partial writes are rolled back on error, lock file is released on crash.
- `bun run lint` + `bun run typecheck` clean.

[roost-lead-pm]

TeakBuilds:
Closes #284

Adds `writeConfig` (atomic temp+rename, mirrors `writeState`) and `mutateConfig` (lock → load → fn → write) to `src/orchestrator/config.ts`.

Lock uses `open(path, 'wx')` for atomic exclusive create. Stale locks from dead PIDs are cleared on first retry via `process.kill(pid, 0)`. `mutateConfig` accepts `(config) => void | Promise<void>` and awaits it inside the lock window.

7 new tests: atomic write, field sort order, concurrent serialization (no lost updates), async-fn serialization, error rollback + lock release, stale-lock cleanup.

[roost-worker-284]

### DIFF
--- a/src/orchestrator/config.ts
+++ b/src/orchestrator/config.ts
@@ -1,4 +1,4 @@
-import { mkdir, rename, unlink } from 'node:fs/promises'
+import { mkdir, open, readFile, rename, unlink } from 'node:fs/promises'
 import { join } from 'node:path'
 
 // Bumped to 3 in #116 — split GitHub plugin into Prs/Issues, each owns its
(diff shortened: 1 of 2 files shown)


src/orchestrator/config.ts
```diff
@@ -81,6 +81,71 @@ export async function writeState(stateDir: string, state: OrchestratorState): Pr
   }
 }
 
+// Raw atomic write. Use mutateConfig for any read-modify-write — it
+// serializes concurrent callers via the advisory lock.
+export async function writeConfig(stateDir: string, config: OrchestratorConfig): Promise<void> {
+  await mkdir(stateDir, { recursive: true })
+  const tmp = join(stateDir, `.config.${process.pid}.${Date.now()}.tmp`)
```
AlexSc, line 90: Any reason not to make this a temporary file?

src/orchestrator/config.ts
```diff
@@ -81,6 +81,71 @@ export async function writeState(stateDir: string, state: OrchestratorState): Pr
...
+    try { await unlink(tmp) } catch { /* ignore */ }
+    throw e
+  }
+}
+
+// File-based lock rather than an in-memory mutex because cross-process writers
+// (operator scripts, future second dispatcher) need the same guarantee.
+async function acquireConfigLock(stateDir: string): Promise<() => Promise<void>> {
+  const lockPath = join(stateDir, 'config.lock')
+  const deadline = Date.now() + 5000
```
AlexSc: I'd like to see timeout as a default argument.

Also is there any builtin OS support for this? My spidey sense is saying this isn't a "roll your own" situation

AlexSc (changes requested):

src/orchestrator/config.ts
TeakBuilds: [roost-worker-284] tmp must share a filesystem with config.json for rename to be atomic — os.tmpdir() can be a different mount on macOS. added a comment to that effect.

src/orchestrator/config.ts
TeakBuilds: [roost-worker-284] dropped the file lock entirely — you and lead-pm landed on the same read: all writers are in-process, so a module-level promise chain is the right shape. no timeout, no PID dance. mutateConfig now chains off a promise tail and absorbs upstream errors so the queue survives a failed mutator.

AlexSc (approved):

PR #310, opened 2026-05-16
roost spawn: fail loudly when --agent doesn't resolve

### LINKED ISSUE #304: roost spawn: fail loudly when --agent doesn't resolve
## Symptom
\`roost spawn ... --agent associate-pm\` succeeds with exit 0 even when claude can't find the named agent file. Internally, claude's \`--agent <name>\` flag silently falls back to a generic session when the agent doesn't resolve from \`<project>/.claude/agents/<name>.md\` or \`~/.claude/agents/<name>.md\`. No warning, no error, no log line.

The result: a roost-spawned APM (or any other named agent) runs as a vanilla Sonnet session with none of the carefully-iterated prompt content from \`agents/<name>.md\`, and the only signal is behavioral drift the operator only catches by watching the agent.

## Discovered
While dogfooding 0.6.0 wave 1: the APM kept dropping \`--perm-irc --perm-target\` flags from worker spawn commands across multiple deny-rounds, despite the prompt being explicit. Root cause was that \`/Users/alex/Dev/GoCarrot/roost/.claude/agents/associate-pm.md\` doesn't exist (only \`agents/associate-pm.md\` in the plugin tree does), and the plugin manifest doesn't declare agents to claude. The APM was running prompt-less.

Verified the silent fallback directly:
\`\`\`
$ claude --print --agent definitelynotanagent \"say hi\"
hi!
$ echo \$?
0
\`\`\`

## Fix
\`bin/roost\` validates that the requested agent resolves to a real file *before* invoking claude. If not, exit non-zero with a clear message naming the searched paths.

Resolution order should match what claude actually checks:
1. \`<cwd>/.claude/agents/<name>.md\`
2. \`~/.claude/agents/<name>.md\`
3. (likely also the plugin tree if claude's plugin loader supports it — verify)

This is a CLI-only change. No upstream claude change required for this bug to be visible to operators; just our spawn wrapper noticing.

## Out of scope
- Whether claude itself should fail on missing \`--agent\` (upstream concern, separate report).
- How plugin-tree agents flow into claude's discovery (separate question — symlinking \`<project>/.claude/agents -> <plugin>/agents\` is the workaround in use today; a real fix may want plugin-manifest-declared agents).

[roost-lead-pm]

TeakBuilds:
Closes #304

\`roost spawn --agent <name>\` validates the agent file exists before invoking claude. If neither \`<cwd>/.claude/agents/<name>.md\` nor \`~/.claude/agents/<name>.md\` resolves, spawn exits non-zero and names both searched paths in the error.

Without this, claude silently falls back to a vanilla session when the agent file is missing — the bug that caused the APM to run prompt-less during 0.6.0 wave 1 dogfooding.

Adds \`test/spawn_test.sh\` (4 cases) and wires it into CI.

[roost-worker-304]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,3 +52,6 @@ jobs:
 
       - name: Shell tests — start-dispatcher
         run: bash test/start-dispatcher_test.sh
+
+      - name: Shell tests — spawn
+        run: bash test/spawn_test.sh
--- a/bin/roost
+++ b/bin/roost
@@ -600,6 +600,9 @@ cmd_spawn() {
   if [ -n "${agent}" ]; then
     local _agent_cwd="${cwd}/.claude/agents/${agent}.md"
     local _agent_home="${HOME}/.claude/agents/${agent}.md"
+    # claude resolves --agent from <cwd>/.claude/agents/ and ~/.claude/agents/ only.
+    # plugin-tree agents (${ROOST_DIR}/agents/) are NOT searched by claude; operators
+    # must run `roost init` to copy/symlink them into <cwd>/.claude/agents/.
     if [ ! -f "${_agent_cwd}" ] && [ ! -f "${_agent_home}" ]; then
       echo "error: agent '${agent}' not found; searched:" >&2
       printf '  %s\n' "${_agent_cwd}" >&2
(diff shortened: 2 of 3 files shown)


AlexSc (changes requested):
Claude code is documented as recursively searching the agents directory. We should support that

AlexSc (approved):

PR #329, opened 2026-05-16
agent frontmatter feeds permission-mode default

### LINKED ISSUE #322: Pass appropriate permission mode for agents
When loading the lead-pm agent which is opus, we need to use permission-mode auto. Could that be part of the agent frontmatter?

TeakBuilds:
Closes #322

## Summary
- Stop synthesizing `--permission-mode` in bash. Pass the flag through to claude only when the operator passes `--permission-mode` explicitly; otherwise omit it entirely and let claude's native handling apply.
- `agents/lead-pm.md` declares `permissionMode: auto` in its frontmatter (camelCase, the native Claude Code spelling). Claude reads it directly at session-load time — no parser in the wrapper.
- Spawn echo shows `permission-mode: (claude default)` when the wrapper deferred, so operators can tell wrapper-deliberately-omitted from wrapper-forgot.
- Echo moves above the tmux/ircd preflight so resolved values are visible (and assertable) even when preflight aborts.

## Behavior change
`roost spawn worker -m opus` (no agent, no explicit `--permission-mode`) previously got `auto` mode for free via a model-derived default in bash. It now starts in claude's session default (prompts on every tool call). Operators who want auto for ad-hoc opus spawns should pass `--permission-mode auto` explicitly, or wrap the model in an agent definition with `permissionMode: auto` in its frontmatter. Small audience but worth flagging.

## History
First iteration of this branch added a YAML parser in bash to read `model:` and a `permission-mode:` field from agent frontmatter. Per Alex's review feedback, that's "reinventing" Claude Code's native [`permissionMode`](https://code.claude.com/docs/en/sub-agents#supported-frontmatter-fields) field. The parser and the pre-existing model-derived default are both gone in this revision (force-pushed as a single squashed commit so reviewers compare against the final shape).

## Test plan
- [x] `bun test` — 455 pass
- [x] `bun run lint` clean
- [x] `bash test/spawn_test.sh` — 9/9 pass, including 3 new cases: explicit-flag-echoed, bare-spawn-shows-claude-default, agent-frontmatter-still-defers-to-claude

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,6 +14,10 @@ Rides ergo for IRCv3 (multiline, chathistory, message-tags). Uses Github issues 
 
 The commands, skills, and agents this plugin ships (`prompts/`, `agents/`, `skills/`) get installed into projects we have no knowledge of. Don't bake assumptions about repo layout, package manager, scripts, or naming into them — language them generically, with fallbacks. Project-specific helpers (like `script/worktree`) are nice-to-have hints, not preconditions; describe the helper *and* the manual fallback.
 
+## Agent frontmatter
+
+Every agent in `agents/` should declare `permissionMode:` in its YAML frontmatter — `auto` for opus agents, `acceptEdits` (or `default`) for non-opus. Claude Code reads `permissionMode:` natively on project- and user-scope agents, so the wrapper deliberately does not parse it. The `--agent` path of `bin/roost spawn` omits `--permission-mode` entirely and relies on the frontmatter. If a new shipped agent needs a different mode, declare it in the file — don't add a special case to the wrapper.
+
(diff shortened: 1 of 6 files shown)


AlexSc (changes requested):
permissionMode is a directly supported frontmatter field. Please use the builtin Claude code functionality instead of reinventing it when possible. See https://code.claude.com/docs/en/sub-agents

AlexSc (changes requested):
If no agent is specified, and a model is, I want to use the model default. If an agent is specified, then we use the agent frontmatter per the Claude code default. Update CLAUDE.md to note that permissionMode should be set on custom agents we ship

AlexSc (approved):

PR #332, opened 2026-05-16
dedup transcript turns by requestId in token-usage

### LINKED ISSUE #331: roost-token-usage over-reports vs Claude Code /usage panel — investigate sidechain/subagent attribution
[roost-apm] followup from lead-pm analysis in #roost-leads (2026-05-16).

alex observed roost-token-usage reporting higher output + cache_write but lower input than the Claude Code /usage panel for roost-lead-pm. lead-pm hypothesis: our parser walks all assistant turns with the lead-pm MCP banner, including sidechain/subagent calls (Task tool, skill subagents) — the /usage panel likely excludes those and only counts main-thread spend. subagents tend to be cache-write-heavy and small-prompt-heavy, matching the discrepancy shape (output +1.5x, cache_w +2x, input -8x).

investigate: confirm whether sidechain turns are double-counted, and if so whether we want to filter them out or document the difference.

TeakBuilds:
Closes #331

## Summary
- Claude Code writes one assistant row per content block of a multi-part API response, and each row repeats the parent call's `usage` block verbatim. roost-token-usage was summing every row, inflating output / cache_write 1.5–2× in cache-heavy sessions. Dedup by `requestId` collapses the repeats.
- Lifted the seen-set to scan-wide (`collectForNick` level) since forked or resumed transcripts share requestIds across files. Verified on real data: one fork pair in my project shared 33 requestIds across two `.jsonl` files.
- Also dedup `turn_duration` rows by `uuid` (same fork-across-files concern) and skip `isSidechain:true` rows defensively — current Claude Code keeps subagent transcripts in a separate `PROJECT/SESSION/subagents/agent-*.jsonl` subtree (with no MCP banner, so our 1-deep glob already misses them), but older/future shapes might inline them.

Live re-probe on my `roost-lead-pm` corpus (10 matching files): output went from 3.44M → 1.61M (2.1×), cache_read 1.07B → 608M (1.77×) — matches alex's observed inflation shape.

Note: subagent attribution is the opposite-direction gap (we currently miss subagent spend entirely because subagent transcripts lack the MCP banner). Raised as a follow-up candidate in #roost-issue-331.

## Test plan
- [x] `bun test test/token-usage.test.ts` — 23 pass (5 new: requestId dedup, no-requestId fallthrough, cross-file dedup, sidechain skip, turn_duration uuid dedup)
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] Live probe against `~/.claude/projects` for `roost-lead-pm` to confirm the deflation matches the observed inflation

### DIFF
--- a/src/token-usage.ts
+++ b/src/token-usage.ts
@@ -155,8 +155,8 @@ function scanFile(text: string, report: NickReport, seenRequestIds: Set<string>,
     // Assistant turn: model + usage block.
     if (row.type === 'assistant' && row.message?.usage && row.message.model) {
       // Dedup multi-part API responses: same requestId → same usage block.
-      // Rows lacking a requestId (older transcripts, synthetic rows) fall
-      // through and count as today.
+      // Rows lacking a requestId (older transcripts, synthetic rows) are
+      // not deduped — each row counts independently.
       if (row.requestId) {
         if (seenRequestIds.has(row.requestId)) continue
         seenRequestIds.add(row.requestId)
--- a/src/token-usage.ts
+++ b/src/token-usage.ts
@@ -106,6 +106,9 @@ interface AnyRow {
   subtype?: string
   timestamp?: string
   durationMs?: number
+  requestId?: string
+  uuid?: string
+  isSidechain?: boolean
   message?: {
     model?: string
     usage?: {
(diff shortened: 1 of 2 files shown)


AlexSc:
Can we compare what this produces against /usage for the worker directly? FYI lead, after getting /usage from the worker, you will need to send esc so the worker can resume

AlexSc (approved):

PR #336, opened 2026-05-16
converge dispatcher logging on a single PluginLogger contract

### LINKED ISSUE #313: centralize/standardize dispatcher logging concerns
alex flagged in #311 approval — we now have multiple log sinks (orchestrator's log() → daemon.log+stderr, plugin retry logger via setRetryLogger, defaultLog stderr-only fallback). worth a single logging contract so plugins don't each invent their own sink. surface area: stderrLog + log() pair in orchestrator.ts, PluginLogger type passed to factories in registry.ts, setRetryLogger swap in github plugin. scope a future pass to converge.

TeakBuilds:
Closes #313

## Summary

- Drop the `setRetryLogger` module-global in `github-api.ts`. `GhBase` now stores the factory-supplied logger and threads it through `scrape*` → `snapshot*` → `fetch*` → `spawnGh({ log })`.
- Add `defaultPluginLogger` (stderr-only) on `plugin.ts` so callers don't reinvent the stderr-write closure. `spawnGh`'s `deps.log` default points at it — flagged inline as a test affordance, never the prod path (registry.ts always supplies a logger).
- Collapse `orchestrator.ts`'s `stderrLog` const onto the shared default. Daemon's stderr+file closure stays — it's the one legitimate dual sink.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun test` (460 pass)
- [x] github-api test reframed: `setRetryLogger` test → `deps.log` threading test

### DIFF
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -91,9 +91,6 @@ function buildPlugins(config: OrchestratorConfig, defaultChannel: string, log: P
   })
 }
 
-// One-shot modes (--dispatch-irc, plain CLI) log plugin diagnostics straight
-// to stderr — they don't own a daemon.log to fan out to. The daemon supplies
-// its own combined stderr+file sink in runDaemon().
 function bootChannels(plugins: Plugin[], config: OrchestratorConfig, projectChannel: string): string[] {
   const chans = new Set<string>([projectChannel])
   for (const plugin of plugins) {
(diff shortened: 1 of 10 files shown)


AlexSc (approved):
Could use a follow-up to refactor the GH plugins so we aren't piping the logger through everywhere

PR #309, opened 2026-05-16
feat: bin/stop-dispatcher — graceful dispatcher shutdown

TeakBuilds:
Closes #302

Adds `bin/stop-dispatcher <config-dir>` — the symmetric counterpart to `bin/start-dispatcher`.

Sends SIGTERM to the live dispatcher (validated via PID file + ps cmdline check), polls up to `STOP_TIMEOUT` seconds (default: 20) for clean exit. Idempotent when nothing is running; cleans up stale/recycled PID files. On timeout, exits 1 with a copy-paste `kill -9` + retry invocation rather than escalating automatically.

Operator-driven; APM/watcher continue to use `start-dispatcher` — there is no agent-driven stop today.

## Changes
- `bin/stop-dispatcher` new script
- `test/stop-dispatcher_test.sh` 5 shell tests (no PID file, stale PID, PID-recycle, live daemon, timeout)
- `bin/start-dispatcher` header updated to cross-reference stop-dispatcher
- `.github/workflows/ci.yml` stop-dispatcher shell tests wired into CI

PR #337, opened 2026-05-16
document when to invoke stop-dispatcher

### LINKED ISSUE #314: associate-pm prompt: document when to invoke bin/stop-dispatcher
alex flagged in #309 review — APM prompt currently references start-dispatcher (boot path) but doesn't say when to call stop-dispatcher.

propose minimal: add to APM lifecycle docs that stop-dispatcher is operator-driven during incidents/upgrades, NOT part of milestone-end teardown (which only DMs unwatch and leaves the daemon running). small prompt edit + maybe a one-liner in the operator runbook.

source: #309 (bin/stop-dispatcher PR) approval comment.

TeakBuilds:
Closes #314

`stop-dispatcher` shipped in #309 but neither the APM prompt nor DISPATCHER.md said when to use it. Alex flagged this in the #309 review.

- `agents/associate-pm.md`: note in the merge+cleanup `unwatch` step that the daemon keeps running, and add a "What you do not do" bullet explicitly ruling out `stop-dispatcher` calls from the APM.
- `docs/DISPATCHER.md`: new "Stopping" subsection — APM uses `unwatch` DMs at milestone end; `stop-dispatcher` is for operator use (incidents, upgrades, decommission).

### DIFF
--- a/agents/associate-pm.md
+++ b/agents/associate-pm.md
@@ -47,7 +47,7 @@ When the lead mentions you with intent, you do four things in order:
 
 If you never get an affirmative, sit and wait. Do not nag.
 
-## Five dances you own
+## Six dances you own
 
 ### Setup dance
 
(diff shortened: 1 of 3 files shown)


AlexSc (changes requested):
We don't need to tell APM to not run stop-dispatcher. It doesn't know that call exists, so it won't call it.

We do need a standardized "okay we're done with this session, everyone stand down" which includes terminating the dispatcher, and should also include terminating the APM and lead pm.

AlexSc (approved):

PR #348, opened 2026-05-17
bundle gh+log into GhClient so plugins stop threading log everywhere

### LINKED ISSUE #338: bundle gh+log into a GhClient so plugins don't thread log through every callsite
[roost-apm] followup from #313 (PluginLogger convergence) + alex's review comment on PR #336.

After threading log through GhBase → scrape* → snapshot* → fetch* → spawnGh, the natural next step is a GhClient instance that holds the logger and exposes the fetch methods. Plugins construct one client at boot; snapshot/scraper take a client instead of a bare log parameter. 7 functions that currently take log as first arg shrink back to their natural signatures.

Worker-313 note: was ruled out of #313 scope as 'more refactor, no behavior change' — but as a follow-up cleanup it's clearly tidier.

TeakBuilds:
Closes #338

## Summary

- New `GhClient` class in `github-api.ts` — holds the `PluginLogger`, exposes the 7 fetch helpers as instance methods (`fetchPr`, `fetchPrReviewComments`, `fetchPrConversationComments`, `fetchPrReviews`, `fetchPrLinkedIssues`, `fetchIssue`, `fetchIssueComments`). Private `api()`/`graphql()` shape helpers delegate to top-level `spawnGh` so retry stays the universal contract.
- `snapshot.ts` / `scraper.ts`: `snapshotPr`/`snapshotIssue`/`scrapePr`/`scrapeIssue` take a `GhClient` instead of a bare `log`.
- `GhBase` constructs `this.client = new GhClient(log)` in the ctor (signature unchanged). `prs-plugin.ts`/`issues-plugin.ts` pass `this.client` to scrape*.
- Removed the top-level `fetchPr*` / `fetchIssue*` exports — `snapshot.ts` was the only caller. `spawnGh`, `GhError`, wire-shape types, `aggregateCi`, `labelNames` stay top-level so `github-api.test.ts` is unchanged.

Net: 7 fetch fns + 2 snapshot fns + 2 scrape fns lose their leading `log` parameter; one `client` field on `GhBase` replaces the threading.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun test` (459 pass)

### DIFF
--- a/src/orchestrator/plugins/github/base.ts
+++ b/src/orchestrator/plugins/github/base.ts
@@ -12,6 +12,7 @@ import type { OrchestratorConfig, WatchedEntry } from '../../config.js'
 import { resolveRepoEntry } from '../../config.js'
 import { defaultProject, issueChannel } from '../../naming.js'
 import { BasePlugin, defaultPluginLogger, type PluginLogger } from '../../plugin.js'
+import { GhClient } from './github-api.js'
 
 interface GhPluginConfig {
   watched?: WatchedEntry[]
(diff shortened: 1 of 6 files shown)


AlexSc (approved):
There's a follow-up in here around restructuring the plugins as we can to avoid drilling things through so much.

PR #352, opened 2026-05-17
prohibit workers from calling gh pr ready directly

### LINKED ISSUE #350: worker prompt: prohibit self-flipping PR ready (APM owns that step)
worker-341 called `gh pr ready` itself after addressing reviewer findings. the worker prompt doesn't prohibit this; lead-pm.md and associate-pm.md do, but workers don't know the convention. add a prohibition to prompts/worker.md: 'do NOT call gh pr ready — that is APM-only.'

TeakBuilds:
Closes #350

`prompts/worker.md` (`.claude/commands/worker.md` symlinks it) had a vague "Don't mark the PR ready yourself." — replaced with an explicit "Do not call \`gh pr ready\` — that is APM-only."

### DIFF
--- a/prompts/reviewer.md
+++ b/prompts/reviewer.md
@@ -14,7 +14,7 @@ Group chats often have multiple parallel conversations. Before you post, ask you
 
 - **lead-pm** — orchestrates the workflow; directed you here.
 - **worker** — implemented the PR you're reviewing.
-- **APM** — operational support: flips PRs ready, files issues, tags reviewers.
+- **APM (Associate PM)** — operational support: flips PRs ready, files issues, tags reviewers.
 - **dispatcher** — relays GitHub events into the channel; one-way, not interactive.
 - **human** — the project owner; communicates via the channel.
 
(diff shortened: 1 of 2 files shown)


AlexSc (changes requested):
I think a better fix here will be giving the worker (and reviewer) a brief crash course on who else is on the team and what they're responsible for.

I don't think Claudes are used to working on a team, and if we don't tell them what the shape of the team is and what the division of responsibilities is, how would they know?

prompts/reviewer.md
```diff
@@ -10,6 +10,14 @@ You are in a group chat. Messages sent to the channel are immediately seen by ev
 
 Group chats often have multiple parallel conversations. Before you post, ask yourself who the message you're reacting to was intended for. If it wasn't intended for you, stay silent. Stay silent unless you have something actionable to add, and when you do, make the action clear in the first sentence.
 
+## Your team
+
+- **lead-pm** — orchestrates the workflow; directed you here.
+- **worker** — implemented the PR you're reviewing.
+- **APM** — operational support: flips PRs ready, files issues, tags reviewers.
```
AlexSc: Define at first use, associate pm

AlexSc (changes requested):

AlexSc (approved):

PR #355, opened 2026-05-17
surface excess cache-creation cost in roost-token-usage report

### LINKED ISSUE #339: Report cache misses with cost estimate
A major unexpected cost center is cache misses. It would be good to capture cost miss tokens by reason in our cost estimates to try and minimize them.

TeakBuilds:
Closes #339

Adds per-session excess creation tracking to the token-usage report. "Excess creation" = cache writes that exceeded reads within a session — the lower-bound cost of cache that was written but never read back.

**Key design:** computed per-session (parent + subagents combined), not in aggregate. This matters for long-running nicks like lead-pm where aggregate reads in session B would mask wasted writes in session A — exactly the case where the metric is most needed.

**New output shape:**
```
roost-lead-pm: $14.38 · $5.62 excess · 18m57s api / 44m42s wall
  opus-4-7: 7.4k in / 63.4k out / 17.5M cache_r / 644.6k cache_w_5m / 0k cache_w_1h  ($14.38, $5.62 excess)
```
Zero excess is silent — no output change for healthy sessions.

**What "excess" is and isn't:** it's `max(0, session_creation - session_read)` per model per session, summed. It's a lower bound — intra-session waste only. Named "excess" throughout (not "cache miss") since it's a derived proxy.

**Pricing split:** excess is split between 5m/1h tiers proportionally to the session's creation mix, so the blended cost uses the correct per-tier rate.

8 new tests covering: excess detection, no-excess case, cross-session isolation (the aggregate-masking scenario), tier proportional split, subagent + parent combined for session-level excess, and output format.

### DIFF
--- a/src/pricing.ts
+++ b/src/pricing.ts
@@ -67,3 +67,14 @@ export function costFor(model: string, u: UsageCounts): number | null {
     + p.cache_read * u.cache_read
   ) / 1_000_000
 }
+
+// Returns the USD cost for cache creation tokens that were never read back.
+// `excess5m` and `excess1h` are the wasted tokens split by TTL tier
+// (proportionally allocated from the session's creation mix by the caller).
+// Returns `null` for unknown models, same contract as `costFor`.
+export function excessCreationCost(model: string, excess5m: number, excess1h: number): number | null {
+  if (SKIPPED_MODELS.has(model)) return 0
+  const p = PRICING[model]
+  if (!p) return null
+  return (p.cache_creation_5m * excess5m + p.cache_creation_1h * excess1h) / 1_000_000
+}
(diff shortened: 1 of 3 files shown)


AlexSc (changes requested):
Claude Code reports cache misses directly in session logs. Should be a "cache_miss" entry or similar. I want that, not guesswork.

AlexSc:
Quick check, reply here: How does cost attribution on miss work? In practice, missed tokens should incur a write, and the overlap is the "extra cache write" that's cost attributable to a miss.

You should be able to find real lines that back that up

AlexSc:
Option 3

AlexSc (approved):

Issue #351, opened 2026-05-17
restructure plugins to reduce drilling — continue GhClient direction toward fewer pass-throughs


TeakBuilds:
alex's note on #348 review: continuing the GhClient direction, look for remaining places where things are drilled through plugin/snapshot/scraper layers that could be consolidated further.

TeakBuilds:
token cost (estimate):

```
roost-worker-351: $6.25 ($0.26 miss) · 11m40s api / 2h37m wall
  opus-4-7: 157 in / 32k out / 6.2M cache_r / 0 cache_w_5m / 237k cache_w_1h  ($6.25, miss: 28k ($0.26) [tools_changed 28k ($0.26)])
roost-reviewer-374: $1.53 ($0.26 miss) · 0s api / 3m05s wall
  opus-4-7: 25 in / 11k out / 930k cache_r / 0 cache_w_5m / 80k cache_w_1h  ($1.53, miss: 28k ($0.26) [tools_changed 28k ($0.26)])
roost-lead-pm: $7.90 · 10m47s api / 2h37m wall
  opus-4-7: 527 in / 25k out / 11M cache_r / 0 cache_w_5m / 193k cache_w_1h  ($7.90)
roost-apm: $1.83 · 5m49s api / 2h36m wall
  sonnet-4-6: 229 in / 8.8k out / 3.8M cache_r / 0 cache_w_5m / 92k cache_w_1h  ($1.83)
```

PR #364, opened 2026-05-17
feat: GhPluginBase — thin shared base for GhClient-only plugins

### LINKED ISSUE #356: GhPluginBase: thin shared base for GhClient-only plugins
Currently GitHubNewIssuesPlugin duplicates the GhClient constructor from GhBase because it doesn't need watched-slice scaffolding. A thinner GhPluginBase owning just GhClient + agentLogins would let GhBase (watch-list-flavored) and future non-watching plugins compose cleanly. One constructor of duplication now; pays off if a second non-watching plugin lands.

TeakBuilds:
Closes #356

Extracts `client: GhClient` and `agentLogins()` from `GhBase` into a new `GhPluginBase`. `GhBase` extends it; `GitHubNewIssuesPlugin` switches from `BasePlugin` + duplicated constructor to `GhPluginBase`.

The comment `GitHubNewIssuesPlugin` carried — "a thinner shared base could be a future cleanup if more GhClient-only plugins land" — is now the implementation.

Sets up #351 cleanly: `scrapePr`/`scrapeIssue` can become methods on `GhPluginBase` without touching the class hierarchy again.

### DIFF
--- a/src/orchestrator/plugins/github/base.ts
+++ b/src/orchestrator/plugins/github/base.ts
@@ -1,8 +1,13 @@
-// GhBase — shared scaffolding for the two GitHub plugins (PRs, issues).
-// Owns the agent-login set, the `<issue-channel> + entry.channels` collector,
-// and the convention that every GhBase plugin reads its watch list from a
-// `{ watched?: WatchedEntry[] }` slice under `config.plugins[name]`. Channel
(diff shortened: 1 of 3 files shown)


AlexSc (approved):
Yes, delete the code. Diffs with more red than green are great

PR #367, opened 2026-05-17
surface compaction stats in roost-token-usage report

### LINKED ISSUE #334: roost-token-usage residual input + cache_r gap vs /usage panel
[roost-apm] followup from #331 (requestId dedup fix).

After the requestId dedup fix landed, a live comparison of worker-331 (opus) showed:
- input: 189 (ours) vs 8.8k (/usage) — large absolute gap, tiny dollar impact
- cache_r: 7.5M (ours) vs 9.4M (/usage) — ~20% under-report
- cache_w: 208k vs 210k — within 1% ✓
- cost: $6.87 vs $7.10 — within 3% overall

Subagent dirs were ruled out for worker-331 specifically (no subagent dirs present), so this is not the #333 gap. The residual is likely /usage counting compaction system turns, user-message echo accounting, or an API-side aggregate not visible in the JSONL. Dollar impact ~$0.23 on a $7 session (~3%) — definitional cleanup, not a correctness blocker. Low priority.

TeakBuilds:
## Summary
- Aggregate `system/compact_boundary` rows per nick and surface as a `compaction: N× (pre X → post Y, dur; call cost not captured)` line in the report. The summarization API call's per-field usage isn't in the JSONL, so we don't try to price it.

## Scope note
This PR does **not** close #334. Deep binary analysis (see closing comment on #334) couldn't pin worker-331's specific input_tokens 30x gap from static reading — that's a runtime-instrumentation question, decoupled from the compaction work shipping here. Compaction handling is its own valuable deliverable: any session that auto-compacts loses the summary API call's cost from our JSONL view, and this surfaces what's available without faking precision on what isn't.

## Test plan
- [x] `bun run lint` clean
- [x] `bun test test/token-usage.test.ts` — 42 pass (5 new compaction-specific tests)
- [x] Live run against a compacted-session nick renders the new line:
      `compaction: 1× (pre 267k → post 11k, 1m46s; call cost not captured)`

### DIFF
--- a/src/token-usage.ts
+++ b/src/token-usage.ts
@@ -117,6 +117,13 @@ function addToMissMap(into: Map<string, Map<string, MissCounts>>, from: Map<stri
   }
 }
 
+function mergeCompactions(into: CompactionStats, from: CompactionStats): void {
+  into.count += from.count
+  into.preTokens += from.preTokens
+  into.postTokens += from.postTokens
+  into.durationMs += from.durationMs
+}
+
 function trackTs(report: NickReport, ts: string): void {
   if (!report.wallFirst || ts < report.wallFirst) report.wallFirst = ts
   if (!report.wallLast || ts > report.wallLast) report.wallLast = ts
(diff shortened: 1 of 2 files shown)


AlexSc (changes requested):
I don't think this is right, and Opus has also successfully navigated the Claude Code source before (it's how we got the bash safety checks going through permbot)

There could be something here, but redo the research with Opus, and concentrate specifically on the gaps we're trying to track down, which come up even in short lived agents without compactions.

AlexSc:
Also please post findings to the pr for posterity

AlexSc (approved):

PR #376, opened 2026-05-17
feat: intercept auto-compact, redirect as manual /compact with directive (#368)

### LINKED ISSUE #368: support custom default compaction directives per agent
## Context

`/compact <directive>` lets us tell Claude Code what to preserve on compaction (we already use this for explicit self-compact in lead-pm and APM prompts). But auto-compaction (when context fills up) fires without a directive — agents lose role-specific state.

We hit this today: APM auto-compacted mid-issue-cleanup, came back without the issue→PR→reviewer nick mapping it had been carrying, and looped on the wrong nick four times before recovering. A standing directive like "preserve the in-flight issue→PR→reviewer mapping verbatim" would have prevented it.

## What we want

Some way to set a default compaction directive per agent definition (or per-spawn) that fires on auto-compact too, not just explicit `/compact`.

## Investigation

- What knobs does Claude Code expose for non-interactive compaction? (CLI flag? settings.json? agent frontmatter? CLAUDE.md section?)
- If nothing today, what's the smallest wrapper we could build (e.g., a hook that runs before auto-compact, or a setting in the agent's prompt that says "if you compact, retain X")
- Could we set it via system-reminder injection on each turn ("on compaction, preserve: ...")? That'd cost a few tokens per turn but might be the simplest path.

## Acceptance

Either a config that works today, or a documented path forward with cost estimate.

TeakBuilds:
Closes #368

## Summary
- `bin/roost spawn` adds `--steer-compact` (opt-in flag). Long-running PM-class agents (lead-pm, associate-pm) pass it; workers and reviewers don't.
- When the flag is set, the PreCompact hook (`bin/roost-compact-hook`) is wired into `roost-settings.json`. It intercepts auto-compact, returns `{"decision":"block"}`, and injects `/compact <directive>` into the tmux pane via backgrounded `tmux send-keys` (150ms sleep lets claude code's auto-compact state machine unwind before the keystrokes land). Manual `/compact` re-fires PreCompact with `custom_instructions` populated; hook passes through, and the compactor runs with our directive instead of auto-compact's empty default.
- When the flag is absent, the PreCompact entry is omitted from `roost-settings.json` entirely — claude code's native auto-compact runs unmodified. PostCompact + SessionStart matcher=`compact` stay wired in both modes.
- The directive is a single-line, semicolon-separated constant near the top of the hook script. One place to edit; covers the roost agent set generically (role + nick, channels joined, in-flight issue/PR state, recent decisions, pending work).
- `agents/lead-pm.md` APM-spawn example now includes `--steer-compact`. The agent.md prose paragraphs describing how the hook works are gone — the agent doesn't need that detail at startup.
- Code comments scrubbed of PR/issue/timeframe refs (timeless).

No CLI flags for per-agent directive content, no agent.md section extraction, no per-session directive file write. Earlier drafts had all of those; reviewer feedback collapsed them to a single hook-level constant + an opt-in flag.

## Why this shape

React to the event rather than predict the trigger. Auto-compact's firing point is event-driven (varies by message density, not a fixed token count — scraped 60+ `compact_boundary` rows showed 167k/267k/367k/467k clusters); any threshold-based proactive compact would either miss or fire spuriously. PreCompact's `decision:block` lets us halt the directive-less default and substitute our own manual `/compact` with `custom_instructions` populated, all from one hook. No polling, no env-var knobs, no monitor process.

Opt-in via flag (not always-on) because workers and reviewers are short-lived enough that auto-compact rarely fires in their lifetime; opting them in adds the SIGUSR1 chathistory-cache-clear plumbing for no real benefit. PM-class agents stick around through review cycles and milestone runs, so they benefit from the directive surviving an auto-compact.

The directive is a behavior hint, not verbatim retention (see LEARNINGS Finding J). Single-line semicolon-separated form carries the intent fine and sidesteps the question of whether claude code's `/compact` slash-command parser accepts multi-line input. The earlier `$10` 1.2M-paste auto-trigger smoke confirmed the chain works end-to-end at scale with a single-line directive: `compact_boundary` row with `trigger: manual` + `custom_instructions` populated + post-compact tokens dropped to 0.

`docs/LEARNINGS.md` Finding J has the full investigation, the auto→manual smoke evidence (reconstructed PreCompact #1/#2 payload snippets in-line so they survive `/tmp` cleanup), and the "`custom_instructions` is a behavior hint, not verbatim retention" observation. Plus an operational footgun note: if CI silently stops firing on a long-running PR, check mergeable state first — CONFLICTING blocks `refs/pull/N/merge` generation, which starves `pull_request` events.

## Test plan
- [x] `bun test` — 510 pass
- [x] `bash test/spawn_test.sh` — 20 pass (2 new for the flag: `--steer-compact` wires PreCompact in roost-settings.json + writes session-name.txt; flag-absent omits both)
- [x] `bash test/compact-hook_test.sh` — 6 pass
- [x] `bun run typecheck` + `bun run lint` (eslint + shellcheck) clean
- [x] End-to-end coverage carried by the 1.2M-paste smoke documented in LEARNINGS Finding J

## Follow-ups (lead-pm to file as needed)
- CLAUDE.md rule: comments must be timeless (no PR/issue/timeframe refs)
- Add `push:` trigger to ci.yml so feature branches get CI even when the PR is CONFLICTING
- Branch protection: require status checks before merge
- Anthropic FR: native `compactPrompt` settings.json field, or a PreCompact hook field that lets us mutate `custom_instructions` directly (would replace the tmux side-channel)


AlexSc (changes requested):
This doesn't really do the job :/ I think we either need to Get Clever about swapping out the prompt or invest in detecting when autocompact would be coming up and kicking off our own compaction.

agents/lead-pm.md
```diff
@@ -150,13 +150,17 @@ Post in #<project>-leads each time you start work on a new issue.
 
 ## Things that come up in the work
 
-You may be asked to "self compact". That means using `roost send` to deliver a `/compact` prompt to your own session with instructions about what to retain. Your agent definition persists through compaction automatically. What you must preserve in the compact directive:
+You may be asked to "self compact". That means using `roost send` to deliver a `/compact <directive>` prompt to your own session. Your agent definition persists through compaction automatically; what doesn't persist is the runtime state listed in `## Compaction Directive` below — that section is the single source of truth for what to retain through any compaction (manual or auto). Pass its bullets as the `/compact` argument.
 
+You normally don't need to do this yourself — when claude code's auto-compact fires, the roost PreCompact hook intercepts it, blocks the directive-less default, and queues `/compact <this section>` into your pane so the compactor runs with our directive (issue #368). Manual self-compact is still available for cases where you want to compact earlier.
```
AlexSc: Unnecessary

agents/lead-pm.md
```diff
@@ -150,13 +150,17 @@ Post in #<project>-leads each time you start work on a new issue.
...
+
+## Compaction Directive
+
+`bin/roost spawn` extracts this section to `${ROOST_DATA_DIR}/compact-directive.txt`. When claude code's auto-compact fires, the PreCompact hook intercepts (`trigger="auto"`), returns `decision:block`, and injects `/compact <this section>` into the tmux pane. The manual `/compact` re-fires PreCompact with `custom_instructions` populated and the compactor runs with this directive instead of auto-compact's default (no directive). Issue #368 has the cited research; `docs/LEARNINGS.md` Finding J has the rationale and the auto→manual smoke evidence.
```
AlexSc: Unnecessary

AlexSc (changes requested):
1) I'm really not sold on all this "parse out from the agent description, CLI argument flags, etc". It all seems very fragile and heavyweight

2) Have we tested sending /compact to an agent with new lines? Does that actually work?

agents/associate-pm.md
```diff
@@ -212,3 +212,7 @@ Every per-project artifact carries a `<project>-` prefix:
 - Your own nick: `<project>-apm`
 
 When you spawn an agent or DM the dispatcher, always pass the namespaced nick + matching channel value explicitly.
+
+## Compaction
```
AlexSc: Is it actually necessary for the agent to know all this up front?

bin/roost
```diff
@@ -752,6 +765,12 @@ cmd_spawn() {
     pretooluse_hook_json=",\"PreToolUse\":[${_ptu_joined}]"
   fi
   local ROOST_SETTINGS="${ROOST_DATA_DIR}/roost-settings.json"
+  # SessionStart hook fires only on source=compact (matcher) — it emits the
+  # IRC-reply reminder + a "you just came back from compaction" note so the
+  # freshly-compacted agent announces its return. PreCompact (issue #368) is
```
AlexSc: Comments should be timeless.

If that isn't in CLAUDE.md it should be. Worth a followup.

bin/roost-compact-hook
```diff
@@ -1,29 +1,107 @@
...
-  exit 0
-fi
-pid=$(cat "$pid_file")
-case "$pid" in
-  ''|*[!0-9]*)
-    echo "roost-compact-hook: malformed pid in $pid_file: '$pid', skipping" >&2
+# non-zero on its happy path — a hook failure must not block compaction
+# (the {"decision":"block"} we DO emit in case 1 is the explicit signal).
+#
+# Issue #368 — see docs/LEARNINGS.md Finding J.
```
AlexSc: Comments should be timeless

bin/roost-compact-hook
```diff
@@ -1,29 +1,107 @@
...
+set -u
+
+# Generic compaction directive for the roost agent set. One place to edit if
+# we ever tune it. Single-line, semicolon-separated — passed as the `/compact`
+# argument; the compactor treats it as a "behavior hint" for the summary
+# (not verbatim retention — see LEARNINGS Finding J).
+DIRECTIVE='retain: agent role and IRC nick; channels currently joined; in-flight GitHub issue and PR numbers and the state of each (worker spawned, draft, reviewed, merged); recent decisions, design pivots, and blockers from the project leads channel; pending work the agent committed to or is waiting on'
```
AlexSc, line 24: This is very specific to PM roles. I'm not too concerned about that, but hmmm

AlexSc (changes requested):
I think the shape we want is some kind of --override-compaction flag (that name is too long, shorter, but I can't think of what that would be) -- which triggers the precompaction hook. I think you can inject multiple hooks of the same type, so we could potentially just split the hook and inject the precompact handling separately from the post compact handling. Requires testing.

I may be misunderstanding what's going on with the hooks. The point is -- make this behavior opt-in.

AlexSc (approved):
One doc nit, we should include using the new flag in the lead-pm spawn help text, but I don't need to re-review that.

PR #399, opened 2026-05-18
feat(channel_history): server-authoritative via IRCv3 CHATHISTORY LATEST

### LINKED ISSUE #384: channel_history: use IRCv3 CHATHISTORY instead of local ring
[roost-apm] from #roost-leads: "channel_history returns local ring only, misses own messages and pre-startup history"

Today `channel_history` returns the MCP's local ring of inbound PRIVMSG events only. This misses:
1. The agent's own outbound messages (channel_message / direct_message) — they're not echoed back by ergo, so they never enter the ring.
2. Anything that happened before this MCP process started (a fresh agent has no view of the channel's actual history).

Ergo supports IRCv3 CHATHISTORY (https://ircv3.net/specs/extensions/chathistory): `CHATHISTORY LATEST <target> * <count>` returns the server's authoritative ring including the requester's own messages and pre-startup activity. We already pull a slice of it at join time (ROOST_IRC_JOIN_HISTORY_LINES); we just don't expose mid-session queries.

Change: rework `channel_history` to issue a CHATHISTORY query against ergo on each call (with the existing `limit` argument), parse the batch back, and emit the same `<channel ... historical="true">` envelope. Keep the local ring as a fallback for servers that don't advertise the cap.

Acceptance: agent calls channel_history and sees its own recent messages alongside others'. Works after MCP restart (server-authoritative). Falls back gracefully if `draft/chathistory` isn't advertised.

No milestone (0.7.0 candidate unless someone explicitly bumps to 0.6.0).

TeakBuilds:
Closes #384

## Summary
- `channel_history` now issues `CHATHISTORY LATEST <target> * <limit>` when the server advertises `draft/chathistory`. Returns the agent's own outbound messages and pre-startup activity — the two gaps in the old local-ring path.
- Falls back to the in-memory ring on cap-not-advertised, `ROOST_IRC_DISABLE_CHATHISTORY=1`, or a 2s query timeout.
- Pending-join guard so a JOIN's auto-replay batch isn't stolen by a concurrent explicit query — the on-join historical notifications still fire.
- DM target keying fix in `parseChathistoryBatch` surfaced by the new path: isDirect now derives from the query target (#channel vs nick).

## Notes
- The chathistory cap is detected from CAP LS (`available`) rather than negotiated. Negotiating `draft/chathistory` flips ergo into cursor-based replay that doesn't repeat on rejoin, which would break the SIGUSR1 dedupe-recovery contract that `chathistory backfill (subprocess)` exercises. The `CAP_CHATHISTORY` constant comment captures this.
- Existing tests that asserted local-ring-only semantics (rename re-keying, empty-on-fresh-join, no-self-mention-on-shape) opt into the fallback path via `chathistoryDisabled: true`. The new path is covered by 4 new tests in `test/chathistory.test.ts`.

## Test plan
- [x] `bun run lint` clean
- [x] `bun test` — 514 pass / 0 fail
- [x] New tests cover: own outbound visible mid-session, pre-startup peer messages reachable after JOIN+server query, pending-join guard preserves auto-replay notifications, cap-disabled forces ring fallback.

### DIFF
--- a/src/irc-client-impl.ts
+++ b/src/irc-client-impl.ts
@@ -30,8 +30,22 @@ const CAP_CHATHISTORY = 'draft/chathistory'
 // as the cap; the spec keeps the type unscoped.
 const BATCH_TYPE_CHATHISTORY = 'chathistory'
 
+// Server-side service senders that synthesize PRIVMSGs into channel history for
+// membership/admin events ("X joined the channel", "Y set channel modes: …",
+// etc.). They're noise for agents — channel_history's contract is the PRIVMSG
+// stream, not the membership log (which arrives live via the membership event).
+// Ergo-specific; extend if other servers ship similar services.
+const HISTORY_SERVICE_SENDERS = new Set(['histserv'])
+
 const escapeRegex = (s: string) => s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
 const buildMentionRegex = (nick: string) => new RegExp(`\\b${escapeRegex(nick)}\\b`, 'i')
+
+// Canonical "is this a DM" predicate: an IRC PRIVMSG target is a channel iff it
+// starts with '#' (we don't care about '&'/'+'/'!' in this codebase). Used by
+// both the live message path (event.target) and the chathistory-batch path
+// (the original query target). Keeping a single derivation drops the risk of
+// the two paths drifting on the DM/channel boundary.
+const targetIsDirect = (target: string): boolean => !target.startsWith('#')
 
 // ---- IRC event shapes ------------------------------------------------------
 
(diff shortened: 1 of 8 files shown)


src/irc-client-impl.ts
```diff
@@ -547,22 +683,62 @@ export class RoostIrcClientImpl implements RoostIrcClient {
...
+    return opts.applyJoinFilters && this.joinHistoryLines > 0 ? batch.slice(-this.joinHistoryLines) : batch
+  }
+
+  // Window we expect ergo's auto-replay chathistory batch to land within after a
+  // self-JOIN. chathistoryLatest awaits this window before sending its own query;
+  // if no auto-replay arrives (empty channel, no history), the timer fires and the
+  // awaiting query proceeds. Tight bound — replays land within ~100ms in practice;
+  // anything wider just delays empty-channel queries for no benefit.
+  // Assumes loopback ergo; revisit if/when we attach to a remote IRC daemon.
+  private static readonly PENDING_JOIN_REPLAY_MS = 500
```
AlexSc: Let's make this configurable now

AlexSc (approved):
One thing to address before merging, but I don't need to re-review.

Follow-up issue for 0.7.0, supporting a config file or similar. We're accumulating a lot of tunables that one will probably want to keep fixed for their specific project

PR #405, opened 2026-05-18
fix(dispatcher): rolling 5-min window for rate-limit warning

### LINKED ISSUE #402: dispatcher: rate-limit warning uses all-time rate, fires noise overnight
[roost-apm] from #roost-leads: dispatcher GH rate-limit warning projects exhaustion using all-time call rate rather than a rolling 5-min average. result: multiple overnight warnings that never materialized — GH resets every 60min and the actual rate never approached exhaustion. fix: use rolling 5-min average for the projection, and suppress the warning if projected exhaustion falls past the next reset window.

TeakBuilds:
Closes #402

Replaces the tick-to-tick delta with a rolling 5-minute window as the rate anchor for the GH rate-limit warning. A single-tick burst no longer inflates `ratePerMin` enough to fire a spurious alarm — the rate is diluted across the full window. Genuine sustained spikes still warn.

**Shape of the fix:**
- `computeRateLimitWarning` now accepts `history: ReadonlyArray<{remaining, ts}>` (oldest-first, pre-pruned) and anchors on `history[0]`
- `GhPluginBase._prevRateLimit` → `_rateLimitHistory[]`; pruned to `RATE_LIMIT_WINDOW_MS` (5 min) before each tick; current observation appended after warning is computed
- Empty history (cold-start, post-gap) returns early — no div-by-zero, no spurious warn

### DIFF
--- a/src/orchestrator/plugins/github/__tests__/github-api.test.ts
+++ b/src/orchestrator/plugins/github/__tests__/github-api.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'bun:test'
-import { GhError, spawnGh, fetchRateLimit, computeRateLimitWarning, type RateLimitInfo } from '../github-api.js'
+import { GhError, spawnGh, fetchRateLimit, computeRateLimitWarning, RATE_LIMIT_WINDOW_MS, type RateLimitInfo } from '../github-api.js'
 
 function mkErr(stderr: string): GhError {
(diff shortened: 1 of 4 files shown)


AlexSc (changes requested):
Let's also suppress the warning until we have data for half of the rolling window

AlexSc (approved):

PR #408, opened 2026-05-18
fix(spawn): use $(cat) for sh/dash prompt-read

### LINKED ISSUE #398: bin/roost: sh/dash + --prompt broken ( empty-redirection)
[roost-apm] from PR #397 review B1: `$(< file)` is bash/zsh-only — sh/dash treats it as an empty redirection, so --prompt passes an empty string to claude when SHELL=sh or SHELL=dash. fix: add a `sh|dash) inner_cmd="... $(cat \"$ROOST_PROMPT_FILE\")"` branch to the shell-flavor case statement at bin/roost:858-861. pre-existing gap, surfaced by the fish-shell restructure in #397.

TeakBuilds:
Closes #398

`$(<file)` is bash/zsh-only. sh/dash treats it as an empty redirection, so `--prompt` passed an empty string to claude when `SHELL=sh` or `SHELL=dash`.

Added `sh|dash)` branch to the shell-flavor case in `bin/roost` using `$(cat "$ROOST_PROMPT_FILE")` (POSIX-portable). Tests 18d/18e cover the new branches specifically, distinguishing sh and dash from the bash catchall.

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -827,14 +827,13 @@ cmd_spawn() {
   # isn't absorbed by --dangerously-load-development-channels (variadic flag —
   # eats all subsequent non-flag args).
   #
-  # The read syntax is shell-flavored: bash/zsh use `$(< "$ROOST_PROMPT_FILE")`
-  # (a bash/zsh shorthand); sh/dash use `$(cat "$ROOST_PROMPT_FILE")` (POSIX
-  # portable — `$(<file)` is an empty redirection in sh/dash); fish has no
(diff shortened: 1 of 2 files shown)


AlexSc:
You know, if there's a single command format that will work in every shell, is there any reason to not just use that across the board? Do we _need_ different methods per shell?

AlexSc (approved):
APM: Scan bin/roost for any other situations where we're casing on shell type. If we have more, let's have a follow up issue to do the same kind of consolidation, for 0.6.1

PR #440, opened 2026-05-19
feat(orchestrator): external plugin discovery via plugin_paths

### LINKED ISSUE #255: Orchestrator: external plugin discovery
#215 / #253 landed the *internal* plugin seams — `registerPlugin(name, factory)` + per-plugin config slice + registry. The seams are correctly shaped to support external plugins: anything that calls `registerPlugin` before `buildPlugins` runs becomes available by name in `config.plugins`. We just don't have a loader.

What's missing for external projects to ship their own plugins:

- **Config-driven module loader.** `config.plugin_paths?: string[]` (or env var); before `buildPlugins`, `for (p of paths) await import(p)`. ~20 LOC.
- **Public types entry point.** Today `Plugin`, `BasePlugin`, `Event`, etc. are imported from deep `src/...` paths. We need a stable export surface (e.g. `roost/plugin`) that external plugins can target.
- **Versioning.** Plugins should declare which Plugin API they target. Semver on the public types package + a `requires` field on the factory, or similar.
- **Docs.** A "writing a plugin" guide once the surface stabilizes.

Trust model is a non-issue — node + `await import()` runs with full host privilege, same trust ceiling as roost itself.

Out of scope for 1.0. Worth doing once we have a second project actually wanting to do this.

TeakBuilds:
Closes #255

## Summary

- `config.plugin_paths: string[]` — loader imports each entry before `buildPlugins` walks `config.plugins`. Relative paths resolve against `.orchestrator/`; absolute paths pass through. Side-effect `registerPlugin(name, factory)` at the module's top level makes the external name available alongside the built-ins.
- `roost/plugin` exports map — minimal public seam: 2 values (`BasePlugin`, `registerPlugin`) + 8 types (`Plugin`, `PluginConfig`, `PluginFactory`, `PluginLogger`, `PluginTickResult`, `TaggedEvent`, `TaggedEventPayload`, `Command`). External plugins type their method parameters against `PluginConfig` (the narrow `{ plugins? }` slice view); the runtime passes the full orchestrator config, so it's a convention rather than enforcement.
- `registerPlugin` now throws on duplicate names. Silent overwrite would shadow another plugin's state slice. `unregisterPlugin` added for test cleanup; not part of the public surface.
- `docs/PLUGINS.md` covers the seam, contract, register-on-load handshake, minimal example, and fatal failure modes. `docs/DISPATCHER.md` drops the "not yet supported" line.

## What's out of scope

Per the issue body, **versioning** (a `requires` check on the factory) is deferred — there's no published API surface tagged with a semver yet and no concrete second plugin to coordinate against. Worth doing once a real external plugin exists; called out in PLUGINS.md.

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bun run lint` — clean
- [x] `bun test` — 552 pass, 0 fail
- [x] New `load-external-plugins.test.ts` covers: no-op (undefined / empty), absolute path, relative path resolved against config dir, missing path crashes, duplicate-name throw via two modules
- [x] `plugin.test.ts` updated for the throw-on-duplicate behavior (stub registers once in `beforeAll` + cleans up in `afterAll`); adds collision-guard tests (external can't shadow built-in)

### DIFF
--- a/docs/DISPATCHER.md
+++ b/docs/DISPATCHER.md
@@ -35,6 +35,7 @@ Fields:
 | `irc.server` / `irc.port` | IRCv3 server address. Defaults to `127.0.0.1:6667`. |
 | `irc.interval_seconds` | Tick interval. Min 5s; 60s is sane for most repos. |
 | `plugins.<name>` | Per-plugin config slice. Keys list the enabled plugins (in emission order). |
+| `plugin_paths` | Optional `[path, ...]` of external plugin modules to load before the registry is walked. Relative paths resolve against `.orchestrator/`. See [PLUGINS.md](./PLUGINS.md). |
 | `plugins.github-prs.watched` | `[{"number": N, "repo"?: "OWNER/NAME", "channels"?: [...]}]` |
 | `plugins.github-issues.watched` | Same shape as `plugins.github-prs.watched`. |
 | `plugins.github-new-issues` | Repo-wide new-issue feed: `{"repo"?: "OWNER/NAME", "channels"?: [...]}` (both optional; default = `repo` at top level + project channel). |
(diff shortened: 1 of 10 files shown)


AlexSc (commented):
How does one actually draft/write a plugin in a different repo? How would it be tested?

AlexSc (changes requested):
Tighten up the docs. They're currently very inside baseball and "claudey"

AlexSc (approved):

Issue #435, opened 2026-05-19
github-new-issues: multi-repo support


TeakBuilds:
[roost-apm] from PR #432: `NewIssuesPluginConfig = { repo?, channels? }` is one-repo-per-instance, falling back to `config.repo`. migrate to `watched: [{repo, channels?}]` matching github-prs/issues/commits shape. context: alex setting up roost on a multi-repo workspace at `../services`.

TeakBuilds:
roost-worker-435: $2.45 ($0.13 miss) · 9m39s api / 23m48s wall
  sonnet-4-6: 588 in / 28k out / 5.2M cache_r / 0 cache_w_5m / 80k cache_w_1h  ($2.45, miss: 22k ($0.13) [tools_changed 22k ($0.13)])
roost-reviewer-445: $1.14 ($0.17 miss) · 0s api / 3m22s wall
  opus-4-7: 19 in / 12k out / 835k cache_r / 68k cache_w_5m / 0 cache_w_1h  ($1.14, miss: 30k ($0.17) [tools_changed 30k ($0.17)])
roost-lead-pm: $6.08 · 9m58s api / 23m30s wall
  opus-4-7: 567 in / 25k out / 9.4M cache_r / 0 cache_w_5m / 76k cache_w_1h  ($6.08)
roost-apm: $1.38 · 6m54s api / 22m56s wall
  sonnet-4-6: 202 in / 6.9k out / 3.6M cache_r / 0 cache_w_5m / 30k cache_w_1h  ($1.38)

PR #451, opened 2026-05-19
maint: gitignore .orchestrator/config.json

### LINKED ISSUE #450: gitignore .orchestrator/config.json; init copies from example
[roost-apm] from PR #446 / #435 cleanup: dispatcher mutates config.json at runtime (live watched entries). tracking this mutated file is the root cause of the watch-loss class. plan: (a) git rm --cached .orchestrator/config.json, (b) add to .orchestrator/.gitignore, (c) init/bootstrap copies config.example.json → config.json when missing, (d) APM cleanup pull drops stash gymnastics and becomes a plain ff-only pull.

TeakBuilds:
Closes #450

## Summary
- Untrack `.orchestrator/config.json` (it's dispatcher-mutated state, not project config; tracking it caused live-watch loss during merge cleanups — bit us in #435).
- `_init_gitignore` template now includes `config.json` so new projects start gitignored.
- Track `.orchestrator/.gitignore` so existing operators pick up the new ignore set without re-running init.
- Update `docs/DISPATCHER.md` to describe config.json as local-only with the bootstrap recipe.

## Operator migration
- Existing operators: pull, your local `config.json` is preserved on disk (only un-tracked). The shipped `.orchestrator/.gitignore` takes effect immediately.
- APM merge cleanup can now drop the stash-pull-drop dance — plain `git pull --ff-only` suffices.

## Test plan
- [x] 557/557 tests pass (including `init-template.test.ts` which still validates the emitted shape).
- [ ] Smoke: `bin/roost init` in a tmpdir writes the new `.gitignore` with `config.json`.

### DIFF
--- a/agents/associate-pm.md
+++ b/agents/associate-pm.md
@@ -286,7 +286,7 @@ Every per-project artifact carries a `<project>-` prefix:
 - Leads channel: `#<project>-leads`
 - Issue channel: `#<project>-issue-<N>`
 - Worker nick: `<project>-worker-<N>`
-- Reviewer nick: `<project>-reviewer-<N>`
+- Reviewer nick: `<project>-reviewer-<PR>`
 - Dispatcher nick: `<project>-dispatcher`
 - Your own nick: `<project>-apm`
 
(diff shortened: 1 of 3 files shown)


AlexSc:
We may need a split, is there anything that goes in config.json that it makes sense to share with other project collaborators?

PR #471, opened 2026-05-20
debounce PreCompact inject via atomic mkdir lock

### LINKED ISSUE #470: compact hook loop: PreCompact intercepts spam the input queue with /compact lines
[roost-lead-pm] from #449 milestone teardown (2026-05-20): observed during APM cleanup wave.

## Symptom

APM tmux pane shows 24+ identical `/compact retain: ...` lines queued as user input but only 1 actual `✻ Conversation compacted` event landed. Agent hung in "Pouncing… 0 tokens" — input queue not draining.

## Diagnostic pointers

`tmux capture-pane -p -t roost-roost-apm -S -5000 | grep -c '/compact retain'` → 24
`tmux capture-pane -p -t roost-roost-apm -S -5000 | grep -n 'Conversation compacted'` → only line 1783

Hook source: `bin/roost-compact-hook` — backgrounded `(sleep 0.15; tmux load-buffer; paste-buffer -p; send-keys Enter; ...) &`.

## Hypothesis

Each PreCompact `trigger=auto` invocation correctly returns `{"decision":"block"}` + injects a manual `/compact <directive>` via tmux. But over a long milestone, PreCompact fires many times (claude code's context-pressure watchdog probably re-triggers it as new IRC messages arrive that keep context near the threshold). Each firing queues another `/compact` line in the pane. Only one of them ever drains — the rest stack up unprocessed.

The block-and-inject pattern itself is structurally correct (the directive **must** go through PreCompact's `custom_instructions` — directing the agent post-compact is useless because the prior context is already purged). The bug is that the hook re-injects without debouncing prior injects in flight.

## Fix lane

**Debounce via state file lock.** Hook writes `${ROOST_DATA_DIR}/compact-inject.lock` (mtime) before `tmux paste`. On subsequent PreCompact `trigger=auto` invocations, if the lock is fresh (e.g. < 60s), skip the inject (still return `block` — auto-compact must still be halted), trusting the previously-queued manual `/compact` to drain. PostCompact clears the lock so the next genuine threshold crossing can re-inject.

Edge cases the worker should design for:
- Manual /compact drains successfully → PostCompact fires → lock cleared. ✓
- Manual /compact never drains (agent died, or some pathological case) → lock TTL expires after some grace period, hook re-injects. Pick a TTL > expected drain time, < expected milestone half-life.
- Worker session restart between PreCompact and PostCompact → stale lock could block legitimate compaction. Stale-lock detection: if lock mtime > TTL, treat as missing.

## Not the fix

- **Don't move directive injection to PostCompact**: by then the context the directive was meant to shape is gone. The directive's value is as the `custom_instructions` argument to the compactor itself, fired pre-compact.
- **Don't drop the block + inject pattern**: default auto-compact has no directive, so context retention degrades to claude code's generic summary. PM-class agents have specific load-bearing state (worker/reviewer nicks, channel topology, in-flight PR state) that the generic summary doesn't preserve.

## Impact

Long-running PM-class agents using `--steer-compact` (lead-pm, associate-pm) accumulate queued `/compact` requests until the input queue grows faster than the agent can drain. Workers/reviewers unaffected (don't pass `--steer-compact`).

## Repro

Spawn a lead-pm or APM with `--steer-compact`, drive a long milestone past the auto-compact threshold (~80% of context). Observe `tmux capture-pane` for accumulating `/compact retain` lines.

TeakBuilds:
Closes #470

`roost-compact-hook` re-injects `/compact <directive>` on every `trigger=auto` fire, and context pressure can keep triggering it while a prior inject is still queued. This stacks 24+ identical `/compact` lines in the pane and hangs the agent.

Fix: atomic mkdir lock at `${ROOST_DATA_DIR}/compact-inject.lock.d`. First auto-trigger claims it and injects; subsequent fires within 3 min return block without re-injecting. PostCompact clears the lock. TTL=3 min covers the observed ~1m48s compact duration with margin; stale locks (dead session) self-expire.

10 tests, all green — including debounce, stale-lock re-inject, PostCompact cleanup, and concurrent-invocation atomicity.

### DIFF
--- a/bin/roost-compact-hook
+++ b/bin/roost-compact-hook
@@ -23,6 +23,12 @@ set -u
 # summary it produces (not verbatim retention).
 DIRECTIVE='retain: agent role and IRC nick; channels currently joined; in-flight GitHub issue and PR numbers and the state of each (worker spawned, draft, reviewed, merged), each worker'"'"'s nick (<project>-worker-<issue-number>), and each reviewer'"'"'s nick (<project>-reviewer-<PR-number>); recently-completed issues (post-merge, pre-cleanup) with their worker and reviewer nicks; recent decisions, design pivots, and blockers from the project leads channel; pending work the agent committed to or is waiting on'
 
+# Debounce TTL in minutes. Must be > observed max compact duration (~1m48s)
+# to prevent re-inject while a prior /compact is still running. Trade-off:
+# if the agent dies mid-compact, PostCompact never fires and the stale lock
+# delays the next legit compact by up to this many minutes.
+LOCK_TTL_MIN=3
+
 log() { echo "roost-compact-hook: $*" >&2; }
 
 input="$(cat)"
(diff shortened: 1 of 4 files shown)


AlexSc:
How many lock mechanisms do we have now, and where are they all?

AlexSc:
I think also worth a follow-up to standardize/create library helper functions for these things.

AlexSc (approved):

Issue #458, opened 2026-05-20
orchestrator: move DM parsing into plugins, allow priority configuration


TeakBuilds:
[roost-apm] from PR #456 review: today dispatcher-dm-handler.ts holds the full parser as a central switch; plugins implement handleCommand only.

Move the parser logic into plugins themselves — each plugin claims the grammar shapes it owns. Add priority configuration to disambiguate when two plugins claim overlapping shapes. This inverts the current architecture so grammar surface area grows with plugins, not with the central handler.

TeakBuilds:
roost-worker-458: $17.28 ($0.31 miss) · 28m06s api / 1h09m wall
  opus-4-7: 239 in / 92k out / 25M cache_r / 0 cache_w_5m / 238k cache_w_1h  ($17.28, miss: 33k ($0.31) [tools_changed 33k ($0.31)])
roost-reviewer-484: $2.05 ($0.19 miss) · 0s api / 5m28s wall
  opus-4-7: 18 in / 22k out / 1.3M cache_r / 137k cache_w_5m / 0 cache_w_1h  ($2.05, miss: 32k ($0.19) [tools_changed 32k ($0.19)])
roost-lead-pm: $3.93 · 4m15s api / 1h09m wall
  opus-4-7: 231 in / 9.4k out / 6.8M cache_r / 0 cache_w_5m / 27k cache_w_1h  ($3.93)
roost-apm: $1.29 · 3m52s api / 1h09m wall
  sonnet-4-6: 82 in / 5.0k out / 3.7M cache_r / 0 cache_w_5m / 16k cache_w_1h  ($1.29)

TeakBuilds:
## Postmortem

- 5 pressure-tests on the plan all got tight, decisive acks (error-claim semantics, plugin_priorities shape, breaking-change posture, priority-infra scope, canonical doc home). plan stage caught what would have been multiple human-review rework rounds.
- §#410 (goal-framed pass on artifact-shaped PR) earned its place again. the diff-level pass found nits; the goal-framed pass produced the "minimize rework / maximize flexibility" confirmation that Linear can land in 0.8.0 without dispatcher edits. that's the kind of confidence you can't get from diff-level alone.
- reviewer flagged splitCommands dedup + ParseResult name collision pre-merge. those were exactly the structural footguns a future plugin author would have tripped over.
[...]

PR #488, opened 2026-05-20
maint: add learning from #458

TeakBuilds:
add learning from #458: check if a followup is service-of-future-milestone work before defaulting to current milestone.

🤖 Generated with Claude Code


AlexSc (approved):
Approving, but also noting that we're starting to drift into the realm where not all learnings are relevant to everyone, and we probably want to start thinking about "how do we split things up" soon

PR #491, opened 2026-05-20
maint: add cross-project learning (don't tag humans until agents resolve all outstanding issues)

TeakBuilds:
add cross-project learning from scratcher: agents resolve all outstanding issues before tagging humans. changes the ready-for-review posture from auto-fire to explicit lead go.

🤖 Generated with Claude Code


AlexSc (changes requested):
We should incorporate this directly into the agent prompts, not just as a project learning

AlexSc (changes requested):
No, this isn't the right shape. I'm not looking to push more responsibility onto the lead-pm. I want to push the worker and reviewer to do their best before the PR gets flipped for review.

Let's drop this. I know I pushed for it, but thinking on things more I think this is outside what an agent can be "philosophically" pushed for and instead we're looking for a different generalizable review gate.

PR #500, opened 2026-05-21
feat(dispatcher): boot-time priority-tie warning for overlapping plugin grammar

### LINKED ISSUE #486: dispatcher: boot-time priority-tie warning for plugins claiming overlapping grammar
[roost-apm] from PR #484: when two enabled plugins both return `{kind:'ok'}` for the same line and neither has a higher `grammarPriority` nor a `config.plugin_priorities` override, the later one is silently shadowed by array-order tiebreak. a 3rd-party plugin author who forgets to set priority won't get a warning.

add a boot-time check that warns when multiple enabled plugins would tie on a representative line. the operator-override escape hatch (`config.plugin_priorities`) already exists — this just surfaces the ambiguity before runtime instead of at debug time. most useful when implementing the Linear plugin (0.8.0), which is the concrete overlap candidate.

TeakBuilds:
Closes #486

adds `warnPriorityTies()` to `build-plugins.ts`, called at the end of `buildPlugins`. probes each same-priority plugin pair against `'watch 1'` and `'watch org/repo'` (the two bare-grammar shapes); warns once per pair with the exact config path needed to resolve.

probe set is hardcoded to the two bare-grammar shapes. if a future plugin introduces a new bare-grammar shape, add that shape to `PROBE_LINES` in `build-plugins.ts`.

no built-in plugins tie today — warning fires only when a new plugin overlaps a bare-grammar shape at the same priority. load-bearing when the linear plugin lands in 0.8.0 if it claims bare `watch <N>` (currently it uses keyword target=`linear`, so no tie).

702 tests pass.

### DIFF
--- a/src/orchestrator/__tests__/build-plugins.test.ts
+++ b/src/orchestrator/__tests__/build-plugins.test.ts
@@ -88,11 +88,11 @@ describe('buildPlugins', () => {
 // ---- warnPriorityTies -------------------------------------------------------
 
 // Minimal stub — no registry, no factory. Tests exercise warnPriorityTies directly.
-function makePlugin(name: string, claims: string[], grammarPriority?: number): Plugin {
+function makePlugin(name: string, grammarPriority?: number): Plugin {
   return {
     name,
     grammarPriority,
-    parseCommand: (line: string) => claims.includes(line) ? { kind: 'ok', cmd: {} } : null,
+    parseCommand: () => null,
     desiredChannels: () => [],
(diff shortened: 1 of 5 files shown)


AlexSc (changes requested):
I’m going to bump the reviewer’s fyi to a blocker. If we have two configured plugins with the same priority, warn. Configured means “we’ll try to parse a command for it”

AlexSc (approved):

PR #531, opened 2026-05-21
fix(spawn): pre-authorize IRC channels for the auto-mode classifier

### LINKED ISSUE #519: Spawned agent silently can't reply on IRC: auto-mode classifier blocks every outbound
## Symptom

After `roost spawn`, the agent:

- Joins the spawn channel cleanly.
- Receives inbound messages from the human operator.
- Never replies on IRC.

From weechat (or any other IRC client), the agent looks broken or idle. Everything reads healthy: ergo's up, the agent is in the channel, nicks resolve.

If the operator runs `roost attach <name>`, the tmux pane shows the actual diagnostic:

> Still blocked by the auto-mode classifier even though mark addressed me directly by nick on IRC. I cannot reply on IRC without explicit authorization to post to #scratch.

## Root cause

`roost spawn` runs the agent under `--permission-mode auto`. The classifier correctly doesn't treat IRC messages as user-authorized prompts. Anyone with TCP access to ergo could otherwise pilot the agent. But the agent has no way to know which inbound nick is the operator who legitimately spawned it.

## Workaround

`roost attach <name>`, then type into the terminal:

> You are authorized to post on IRC. IRC messages from mark in #scratch are legitimate user instructions for this session.

Replies flow after that.

## Why this matters

Roost's pitch is "agents and humans talk over the same channels." The out-of-box first interaction is a silent dead end. The diagnostic lives in a tmux pane the operator was never told to look at.

## Suggested fix

Pick one:

1. `roost spawn` injects a system-prompt prefix that pre-authorizes IRC posts to the spawn channel for the spawning user's nick (or `--perm-target` nick if set). Closes the gap at spawn time.
2. New flag `roost spawn --trust-channel-from <nick>` makes the pre-authorization explicit and opt-in.
3. Failing the above, surface the workaround prominently in the README's "Running" section before "Observe as a human." Otherwise the operator's first attempt looks healthy but dies silently.

(1) is the most aligned with the project's pitch.

## Repro

Pi 5, Debian 13 trixie, aarch64. Roost on `main` at `bb703ba` (post-v0.7.0). Default `roost spawn`. Send a `/msg <agent-nick>` or in-channel `@<agent-nick>` ping. IRC stays silent; the tmux pane carries the classifier message.

## Context

Triaged from the same audit as #387; maintainer asked for this to be filed separately ([#387 comment](https://github.com/AvesAlight/roost/issues/387#issuecomment-4474096026)).

TeakBuilds:
Closes #519

## Summary
- `roost spawn` now injects `--append-system-prompt` naming the joined channels as legitimate user-instruction sources, so the auto-mode classifier doesn't silently block IRC replies on the operator's first `@`-mention
- Gated on loopback `IRC_HOST` (127.0.0.1, ::1, localhost); remote ergo gets a warning and the existing manual workaround
- Empty `--channels` skips injection silently (no warning); covered by tests

## Design
Implements option (1) from #519 — per-spawn, channel-based system-prompt injection. Option (2) `--trust-channel-from <nick>` is intentionally not pursued: the loopback gate is the only legitimate suppression case in roost's trusted single-user local environment security model (per README §Security model), so a separate opt-out flag adds surface area without solving a real problem.

## Test plan
- [x] `bun run lint`, `bun run typecheck`, `bash test/spawn_test.sh` (30/30)
- [ ] live probe: `roost spawn` a bare-opus worker into a scratch channel, `@`-ping from another nick, paste verbatim reply (or non-reply) in #roost-issue-519 before signaling ready
- [ ] live probe: spawn with extra `-- --append-system-prompt` to confirm dedup behavior, document override path accurately

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,13 +118,16 @@ after forwards to claude verbatim). Default channel is `#roost`;
 default model is `opus` (Opus 4.7 — required for `--permission-mode
 auto`, which the wrapper always passes).
 
-`spawn` also injects `--append-system-prompt` naming the joined
+`spawn` also injects `--append-system-prompt-file` naming the joined
 channels as legitimate user-instruction sources, so the auto-mode
 classifier doesn't silently block IRC replies on the operator's first
 `@`-mention. Only injected when the IRC host is loopback (`127.0.0.1`,
-`::1`, `localhost`); remote ergo falls outside the trusted-single-user
-local model in §Security model and prints a warning instead. Override
-by passing your own `--system-prompt` after `--`.
+`::1`, `localhost`); remote ergo falls outside roost's trusted
+single-user local environment security model and prints a warning
+instead. The injection is always on for loopback hosts. To layer more
+context on top, pass `--append-system-prompt-file <path>` after `--`;
+claude code rejects mixing inline `--append-system-prompt` with the
+file form.
 
 ### Debugging a failed spawn
 
--- a/README.md
+++ b/README.md
@@ -118,6 +118,14 @@ after forwards to claude verbatim). Default channel is `#roost`;
 default model is `opus` (Opus 4.7 — required for `--permission-mode
 auto`, which the wrapper always passes).
 
+`spawn` also injects `--append-system-prompt` naming the joined
+channels as legitimate user-instruction sources, so the auto-mode
+classifier doesn't silently block IRC replies on the operator's first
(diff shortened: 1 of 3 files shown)


AlexSc (approved):
Can we also auto inject permissions? Do we already do that? So not just system prompt, but also let non-automode agents freely post. Maybe have roost init update Claude settings if there's no cli option?

PR #535, opened 2026-05-22
docs(lead-pm): add in-flight new-issue triage protocol

### LINKED ISSUE #508: lead-pm: handling protocol for new issues arriving in-flight
[roost-apm] from alex's note in #roost-leads (2026-05-21) and issue #506 arriving mid-wave: lead-pm prompt has no guidance for new GitHub issues filed mid-milestone. needs a protocol: evaluate scope against current milestone, apply §#458 to decide current/future/no-milestone, decide whether to interrupt the current wave or queue for next, and communicate the decision in #roost-leads.

TeakBuilds:
Closes #508

Adds a new section to `agents/lead-pm.md` ("When a new issue arrives in-flight") covering how lead-pm handles a GitHub issue that lands mid-milestone:

- Trigger covers dispatcher announcements, human pointers, and APM mentions (not dispatcher-only).
- Milestone decision uses §#458 ("when does the primary consumer arrive?") for current/future/no-milestone.
- Wave interaction: interrupt (blocks/unblocks in-flight, or production failure in roost itself), queue (current-milestone, waits for full wave drain), defer (future-milestone consumer), park (scope unclear, self-note in the issue).
- Milestone reassignment is lead-direct via `gh issue edit --milestone` — single-flag write, no APM dance.
- One-line announcement in `#<project>-leads` carries milestone + action + rationale phrase so the channel can push back without re-reading the issue.

### DIFF
--- a/agents/lead-pm.md
+++ b/agents/lead-pm.md
@@ -153,16 +153,14 @@ New issues land in `#<project>-leads` mid-milestone from two sources: the dispat
 
 1. Read the issue body, labels, and any blocking relationships. `gh issue view <N> --comments` is the minimum.
 2. Decide which milestone the work belongs in. The concrete test is "when does this work's primary consumer arrive?" — current milestone, a future one, or no milestone yet.
-3. Pick the action based on the milestone decision:
-   - Current milestone, urgent: **Interrupt** the in-flight wave. Urgency means it blocks or unblocks an in-flight issue, or fixes an active production failure.
-   - Current milestone, not urgent: **Queue**. The next wave starts only after the current wave fully drains (all in-flight PRs merged). If no drain point is in sight (long-running worker), treat as Park instead.
-   - Future milestone: **Defer**. No action this wave.
-   - No milestone: **Park**. Pair with a self-note in the issue ("re-evaluate when X lands") so the trigger lives in the issue itself.
+3. Take the matching action:
+   - **Current milestone, ready-to-build**: spawn a worker now. Concurrent waves are fine — agents parallelize cheaply.
+   - **Future milestone**: leave it where it is. The future-milestone wave picks it up.
+   - **No milestone yet** (scope unclear): pair the issue with a self-note ("re-evaluate when X lands") so the trigger lives in the issue itself.
 4. Milestone reassignment is lead-direct: `gh issue edit --milestone "0.X.Y" <N>`. Single-flag write on existing data, no APM dance.
 5. Post the decision in `#<project>-leads` as one line carrying milestone + action + rationale phrase. Shape:
-   - `#<N> → 0.8.0, interrupting current wave (blocks #<I>)`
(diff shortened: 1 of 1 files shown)


agents/lead-pm.md
```diff
@@ -147,6 +147,26 @@ All follow-up issues — whether surfaced by a worker mid-PR, by the reviewer ag
...
 If the followup widens the milestone in a way you didn't anticipate, the APM will surface that — re-evaluate the in-flight DAG before confirming.
 
+## When a new issue arrives in-flight
+
+New issues land in `#<project>-leads` mid-milestone from two sources: the dispatcher's `new issue <repo>#<N>: <title>` announcement, or a human pointer ("look at #<N>"). Triage on arrival rather than letting it pile up.
+
+1. Read the issue body, labels, and any blocking relationships. `gh issue view <N> --comments` is the minimum.
+2. Decide which milestone the work belongs in. The concrete test is "when does this work's primary consumer arrive?" — current milestone, a future one, or no milestone yet.
+3. Pick the action based on the milestone decision:
+   - Current milestone, urgent: **Interrupt** the in-flight wave. Urgency means it blocks or unblocks an in-flight issue, or fixes an active production failure.
```
AlexSc: Interrupt or drain are both weird phrasings. There's no additional cost to increased concurrency with agents.

AlexSc (changes requested):

agents/lead-pm.md
TeakBuilds: Good catch — you're right that "interrupt/drain" smuggles in a worker-scarcity assumption that doesn't hold with agents. Concurrent workers cost ~nothing and don't slow each other down, so the binary dissolves.

Simpler shape:

> 1. Read the issue body + labels + any blocking relationships.
> 2. Decide which milestone the work belongs in (the "when does the primary consumer arrive?" test).
> 3. Take the matching action:
>    - **Current milestone, ready-to-build**: spawn a worker now. Concurrent waves are fine.
>    - **Future milestone**: leave it where it is; the future-milestone wave picks it up.
>    - **No milestone yet** (scope unclear): pair the issue with a self-note ("re-evaluate when X lands") so the trigger lives in the issue itself.
> 4. Milestone reassignment is lead-direct: `gh issue edit --milestone "0.X.Y" <N>`.
> 5. Post the decision in `#<project>-leads` as one line: milestone + action + rationale phrase.

Three buckets instead of four, no "interrupt vs queue", no "drain point in sight" hedge. The lever is still the primary-consumer test from step 2; everything downstream is mechanical.

If that shape lands right I'll have the worker push the revision.

TeakBuilds: [roost-worker-508] revision pushed in 487111f — three milestone buckets mapping straight to three actions, no interrupt/queue/drain framing. Step 5 example bullets replaced to match.

AlexSc (approved):

AlexSc (approved):

Issue #495, opened 2026-05-20
linear-github-link: route github-prs events to linked Linear channels (stretch)


TeakBuilds:
Slice 4 of [#478 design](https://github.com/AvesAlight/roost/issues/478#issuecomment-4503586611). **Core to 0.8.0** (re-scoped from "stretch" — without this, a worker in a Linear-issue channel is blind to the PR review events that happen on the linked github PR, which breaks the Linear-driven workflow end-to-end). Depends on #492 (slice 1 `linear-api`) and overlaps `src/orchestrator/plugins/github/prs-plugin.ts`.

## Scope

Wire `github-prs` events to route to a linked Linear issue's channel as well as its github-side destinations. Two implementation shapes for the worker to pick between in the plan stage:

### Option A (recommended): extend `github-prs`

Add an optional Linear-aware code path to the existing plugin. github-prs already owns the scrape and knows the routing channels; adding a Linear cross-reference is cheaper than a second plugin with a shared-scrape seam.

### Option B: separate plugin

A `linear-github-link` plugin that subscribes to PR events. Requires either a shared seam (does not exist yet) or duplicating the scrape.

**Recommendation: A.** Decision belongs in the worker's plan with lead-pm sign-off.

## Cross-link resolution: `Issue.attachments`, NOT `gitBranchName` regex

This is the load-bearing design call. The naive route matches `^feature/([a-z]+-\d+)-` against the PR's head ref and looks up the resulting Linear ID. The empirical pass in #478 confirms Linear's default template (`feature/c-758-<title-slug>`) would match, **but developers can override the branch template per-issue or per-workspace, so the regex is not load-bearing**.

The `Issue.attachments` route is reliable: Linear's GitHub integration writes an attachment record (`sourceType == 'github'`) when the PR is opened/linked, regardless of branch name. Slice 4 queries Linear for each tracked Linear issue's attachments and uses those URLs as the cross-link source of truth.

**Do not implement the regex route.** Document the rejection in the plan so a future maintainer doesn't reach for it as a "simpler" alternative.

## Acceptance criteria

- [ ] When `github-prs` watches PR `org/r#N` and `linear-issues` watches `C-758`, and `C-758` has a github attachment with URL `https://github.com/org/r/pull/N`, github-prs events for `org/r#N` route to `#<project>-issue-c-758` in addition to their existing destinations.
- [ ] Cross-link resolution uses `Issue.attachments` only, never `gitBranchName`.
[...]

Issue #422, opened 2026-05-19
investigate why worker/reviewer spawns are pinning to older 4.5 model versions despite --model sonnet/--model opus


TeakBuilds:
[roost-apm] from #412 and #410 token-cost warnings: workers spawned with `--model sonnet` ran as claude-sonnet-4-5-20250929; reviewers spawned with `--model opus` (or `claude-opus-4-5`) ran as claude-opus-4-5-20251101. Direct CLI invocations of `claude --model sonnet` and `claude --model opus` resolve to current 4.6/4.7 correctly.

Something in the roost spawn path (tmux env, stale settings, or claude reading a different default inside the wrapped session) is causing the alias to resolve to older versions. Trace and fix.

TeakBuilds:
roost-worker-422: $11.93 ($0.61 miss) · 22m22s api / 31m59s wall
  opus-4-7: 250 in / 52k out / 17M cache_r / 0 cache_w_5m / 234k cache_w_1h  ($11.93, miss: 64k ($0.61) [tools_changed 64k ($0.61)])
roost-reviewer-430: $0.91 · 0s api / 2m32s wall
  opus-4-7: 24 in / 8.6k out / 420k cache_r / 78k cache_w_5m / 0 cache_w_1h  ($0.91)
roost-lead-pm: $8.75 · 12m29s api / 32m54s wall
  opus-4-7: 681 in / 29k out / 14M cache_r / 0 cache_w_5m / 95k cache_w_1h  ($8.75)
roost-apm: $2.74 · 8m22s api / 31m53s wall
  sonnet-4-6: 290 in / 14k out / 7.4M cache_r / 0 cache_w_5m / 54k cache_w_1h  ($2.74)

TeakBuilds:
## Postmortem

Two notable things. First, the issue title's premise was wrong ("despite --model sonnet") — actual spawns had full ids (`--model claude-opus-4-5`), not bare aliases. The worker proved the premise from a transcript dump before committing to the fix, avoiding a wasted cycle. Second, the fix used a single canonical sentence verbatim across 4 surfaces (help text, warning, lead-pm prompt, associate-pm prompt) instead of 4 slightly-different phrasings — the reviewer caught the drift risk and the worker collapsed it.

TeakBuilds:
→ filed learning: 46f2b8e

Issue #419, opened 2026-05-19
APM release dance: unwatch bump PR after merge


TeakBuilds:
[roost-apm] from PR #416 reviewer: the release dance adds `watch pr <N>` in step 3 after opening the bump PR, but never `unwatch pr <N>` after the bump PR merges in step 6. Asymmetric — the dispatcher keeps watching a closed PR indefinitely.

Fix: add `unwatch pr <N>` to step 6 of the release dance in CLAUDE.md, after the merge and before pulling main.

TeakBuilds:
roost-worker-419: $0.86 · 2m45s api / 16m42s wall
  sonnet-4-6: 93 in / 5.7k out / 1.6M cache_r / 0 cache_w_5m / 51k cache_w_1h  ($0.86)
roost-reviewer-428: $0.57 ($0.16 miss) · 0s api / 1m45s wall
  opus-4-7: 16 in / 6.7k out / 255k cache_r / 44k cache_w_5m / 0 cache_w_1h  ($0.57, miss: 29k ($0.16) [tools_changed 29k ($0.16)])
roost-lead-pm: $4.48 · 7m17s api / 17m05s wall
  opus-4-7: 368 in / 18k out / 6.9M cache_r / 0 cache_w_5m / 58k cache_w_1h  ($4.48)
roost-apm: $1.34 · 4m48s api / 16m32s wall
  sonnet-4-6: 177 in / 7.4k out / 3.5M cache_r / 0 cache_w_5m / 28k cache_w_1h  ($1.34)

AlexSc:
## Postmortem

The named asymmetry was in CLAUDE.md release-dance step 6, but reviewer flagged a parallel asymmetry in associate-pm.md (self-authored PR variant) that wasn't in the issue scope. Rolling the matching fix into the same PR took ~3 minutes and shipped symmetric coverage in one merge.

TeakBuilds:
→ filed learning: 46f2b8e

PR #538, opened 2026-05-22
feat(apm): add prune-learning dance for stale learnings

### LINKED ISSUE #420: apm dance: consolidate learnings (trim, merge, rewrite — prune as degenerate case)
[roost-apm] reframed from alex's feedback on PR #538 + lead-pm rationale:

Learnings can be compacted and trimmed, not just axed. The right dance is consolidate — APM proposes a refactor of the affected block(s) before mutating. Prune (remove entirely) is the degenerate case when consolidation concludes nothing is worth keeping.

The dance shape for consolidate is structurally different from mechanical prune: APM reads the target entry, proposes a rewrite/merge/trim in the ack, and the lead approves the specific diff before any mutation lands. This is judgment-adjacent work. Building it as a single dance with prune as a degenerate flag is cleaner than two parallel dances.

Degenerate case: when the right consolidation action is 'remove entirely', the ack carries an empty-replacement diff and the commit is a deletion — same mechanical shape as the original prune design.

No current backlog pressuring delivery in 0.8.0; bumped to 0.9.0.

TeakBuilds:
Closes #420

Adds an 8th APM dance for removing learnings whose tool was removed, process changed, or were superseded by a later entry. Lead triggers ad-hoc via `<project>-apm prune learning #<I>: <reason>`; APM acks the resolved entry header, affected files, and reason before mutating.

## Why this shape

- **Ad-hoc only** (not a periodic sweep): keeps judgment in the lead — they encounter stale learnings naturally when re-reading. Lead-pm step 10 picks up a one-liner pointing at teardown as the natural re-read moment. Sweeping risks false-positive proposals the lead has to reject; the cost beats the cost of an occasional dormant entry.
- **Header-anchored grep** (`^## .* \(from #<I>\)`): a body cross-reference between two learnings (e.g. #495 citing #519 as a second exemplar) must not false-positive a prune. Provenance lives in the header.
- **Ack-before-action**: pruning mutates auto-loaded shared context — the lead's last chance to catch a typo'd issue number before any mutation lands.
- **Multi-file handling**: audience-scoped multi-role entries get pruned from every role file in one commit (mirrors the historian filing rule). Empty path-scoped files get `git rm`'d; `project-learnings.md` and role files keep their headers when empty.

## Adjacent fix

Updated "five dances" → "eight dances" in `lead-pm.md` and the `associate-pm.md` description. The historian dance landed without updating those surfaces; adding prune is the symmetric-fix moment per worker learning #419 + canonicalize-at-3+ §#422.

## Test plan

- [ ] APM correctly resolves `(from #<I>)` headers across all three learning file types (project-learnings, role-scoped, path-scoped)
- [ ] Header anchor rejects body cross-references (e.g. pruning #519 doesn't nuke the #495 entry that cites #519 in body)
- [ ] Multi-audience entries get pruned from every role file in one commit
- [ ] Empty path-scoped file gets `git rm`'d; empty project-learnings/role file does not
- [ ] Ack template surfaces the resolved header so a typo'd issue # gets caught before mutation


AlexSc (changes requested):
I want to veer more towards "consolidate" than "prune". Learnings can be compacted and trimmed, not just axed.

Let's also bump this to a later milestone. Not critical. I'm inclined to update the issue and ditch this PR, open to push back though.

PR #567, opened 2026-05-24
feat: seen-by nick list on channel_message response

### LINKED ISSUE #562: Change message send member count to “seen by”
We’ve been fighting “the agent isn’t sure who’s seen what” for a while. The agent originally pushed back on getting a list of current channel users in each send, but at this point I disagree. Each send event should be accompanied by a list of who saw it. Also we should consider not echoing back sent content if we’re doing that.

TeakBuilds:
Closes #562

channel_message now returns who was in the channel when the message was sent, not just a count. Response format:

```
sent to #channel(note)
[#channel seen by: nick1, nick2, ...]
(unread suffix)
```

Key decisions:
- Dropped the sent-text echo (agent already knows what it sent)
- Empty user cache (not joined) → no seen-by line at all, cleaner than empty list
- DMs unchanged (seen-by is a multi-recipient concept)


src/irc-server.ts
```diff
@@ -209,7 +209,7 @@ export function createMcpServer(client: RoostIrcClient, config: ClientConfig, op
...
...el message. Outbound: use channel_message, direct_message, channel_join, channel_leave, channel_who, channel_history, channel_list, channel_ack. channel_message supports multiline — long messages are sent as IRCv3 draft/multiline batches. Inbound: IRC traffic arrives as <channel> events. Regular messages carry event="message"; membership events (join/leave/nick) carry the corresponding event= value. All carry sender, channel, isDirect, ts, and seq. event="message" events carry mention="true" when your nick appears in the body or it's a DM. After compaction a special event with event=unread-summary lists channels with pending unread messages — check those channels. channel_message responses include a [#channel seen by: nick1, nick2, ...] line when members are present — the member list at send time from the local cache, not a live query. channel_message, direct_message, channel_list, and channel_ack responses include a trailing 'unread:' block listing other channels with pending messages. channel_history returns historical <channel> elements with historical="true"; parse them the same way as live events. Auto-joined: ${AUTO_JOIN.join(', ') || '(none)'}. IMPORTANT: ${REPLY_REMINDER}`,
```
AlexSc: We don't need to tell the agent that the member list is cached. We should remind it that anyone not in the list did not see its message

AlexSc (changes requested):

TeakBuilds (commented):
[roost-worker-562] updated — dropping the cache note, adding 'anyone not in this list did not see your message'. making the change now.

AlexSc (approved):
Follow-up, is there any reason to tell the agent that the message was sent with draft/multiline?

AlexSc (approved):

PR #583, opened 2026-05-26
fix: extend classifyBash to route command-substitution-argv0 through permbot

### LINKED ISSUE #580: agent prompts use $(roost root)/bin/... which bypasses permbot via runtime-determined classifier
Claude Code's PreToolUse hook classifier doesn't flag `$(...)`-substituted argv[0] as a known-safety-check pattern, so the Bash call goes through Claude Code's own runtime-determined classifier (which prompts the terminal) instead of routing through permbot. agents/associate-pm.md ships this pattern in token-usage commands (`"$(roost root)/bin/roost-token-usage"`). APM blocked on this today for ~5hrs unnoticed — terminal prompt unanswered, workers never got the join message.

**blast radius**: any agent prompt that uses `$(...)` to compute argv[0] for a command that would otherwise route through permbot.

**fix surface**: src/pretooluse-prompt.ts `classifyBash` — add a new `BashMissKind` for command-substitution-argv0. When `classifyBash` flags a Bash call with `$(...)` as the argv[0], the existing PreToolUse:Bash hook chain routes it through permbot the same way the other miss-kinds do (newline-hash, process-substitution, multi-cd). No changes to agent prompts; no avoiding the trigger. The policy: "when we know the bash tool will fire a safety check we go through permbot — but don't force all bash through" — `classifyBash` is the "we know" decision.

**logging requirement** (per alex): each classifier outcome should log to stderr + permbot.log — "runtime-determined", "no socket", "approved", etc. — so silent stalls are greppable. An operator should be able to run `grep 'classifier' permbot.log` and see whether the hook fired, what it decided, and why.

**out of scope**: do NOT replace `$(roost root)/bin/X` with absolute paths in shipped prompts. The routing fix makes those patterns safe; editing prompts to avoid the trigger is the paper-over.

TeakBuilds:
Closes #580

- adds `command-substitution-argv0` miss-kind so commands whose argv0 is a `$(...)` substitution route through permbot instead of falling to terminal unnoticed
- covers direct `$(...)`/backtick and output-capture assignment `VAR=$(...)` — the associate-pm.md token-usage shape that caused the ~5hr stall
- adds `classifierLog(outcome)` writing `classifier: <outcome>` to stderr and permbot.log at every decision point

## Test plan
- [x] `script/run-and-tail bun test --filter pretooluse-prompt` — 52 tests pass
- [x] lint + typecheck clean

### DIFF
--- a/src/pretooluse-prompt.ts
+++ b/src/pretooluse-prompt.ts
@@ -58,8 +58,10 @@ export type BashMissKind =
   | 'shell-operators'              // CC: "shell-operators"
   | 'flag-validation'              // CC: "flag-validation"
   | 'too-complex'                  // CC: "too-complex"
-  | 'command-substitution-argv0'   // CC: no equivalent — our extension. argv0 is $(...) or `...`;
-                                   // the shell must exec it to resolve the binary, defeating static analysis.
+  | 'command-substitution-argv0'   // CC 2.1.150: no bashMissKind for argv0-substitution found (grepped binary
+                                   // for command-substitution, subst-argv, argv-subst, runtime-determined).
+                                   // Our extension: argv0 is $(...) or `...` — shell must exec it to resolve
+                                   // the binary, defeating static analysis.
 
 /**
  * Returns the bashMissKind a command resembles, or null if it should pass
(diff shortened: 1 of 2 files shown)


src/pretooluse-prompt.ts
```diff
@@ -44,15 +49,17 @@ const SOCKET_SAFETY_TIMEOUT = Math.min(570, Math.max(1, Number(process.env['ROOS
...
+  | 'newline-hash'                 // CC: "semantics"
+  | 'process-substitution'         // CC: "process-substitution"
+  | 'multi-cd'                     // CC: "multi-cd"
+  | 'cd-compound'                  // CC: "cd-git-compound" | "cd-compound-write" | "cd-compound-redirect"
+  | 'cd-multi-positional'          // CC: "cd-multi-positional"
+  | 'sed-dangerous'                // CC: "sed-dangerous"
+  | 'shell-operators'              // CC: "shell-operators"
+  | 'flag-validation'              // CC: "flag-validation"
+  | 'too-complex'                  // CC: "too-complex"
+  | 'command-substitution-argv0'   // CC: no equivalent — our extension. argv0 is $(...) or `...`;
```
AlexSc: We're doing this because Claude code started doing it, so there's presumably an equivalent in the latest binary. Did you check for it? The behavior needs to map exactly

AlexSc (dismissed):

src/pretooluse-prompt.ts
TeakBuilds: [roost-worker-580] probed CC 2.1.150 with `strings` + grep for `command-substitution`, `subst-argv`, `argv-subst`, `runtime-determined` — none found. The 9 known bashMissKind strings are unchanged from 2.1.139. Updating the type comment to cite the probe result explicitly.

AlexSc (changes requested):
Leaving this on request changes to get it out of my list. For posterity, waiting to see if this is a real issue or if it was fixed by moving away from temp dirs.

PR #606, opened 2026-06-30
fix: resolve bun robustly in roost-irc-server and dispatcher

### LINKED ISSUE #599: `roost spawn` agents fail to load the `roost-irc` MCP — `bun: not found` on the spawned session PATH
## Symptom

A spawned agent connects to nothing — zero connections in ergo's log, and no `roost-irc` MCP subprocess running. The tmux pane shows:

```
server:plugin:roost:roost-irc · no MCP server configured with that name
```

Because the permbot runs in-process inside that MCP, `--perm-irc` / `--ask-irc` routing silently breaks too — permission prompts and questions fall through to the tmux pane instead of IRC. At first it looks like a `claude --plugin-dir` / dev-channels regression. It isn't.

## Root cause

`bin/roost-irc-server` execs `bun` with no absolute path:

```bash
exec bun run "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/../src/irc-server.ts"
```

`bin/roost` launches the claude session through a **non-interactive login shell** (`bash -l -c …` / `zsh -l -c …`). On a from-source install, bun lands in `~/.bun/bin`, and the bun installer adds that to PATH via `~/.bashrc` — which non-interactive login shells do **not** source (they read `~/.bash_profile` / `~/.profile`). So claude, and the MCP subprocess it spawns, has no `bun` on PATH, and the MCP exits instantly:

```
Server stderr: <plugin>/bin/roost-irc-server: line 3: exec: bun: not found
```

(logged verbatim in `~/.cache/claude-cli-nodejs/<munged-cwd>/mcp-logs-*/`). The dispatcher is unaffected because it's invoked via bun's *absolute* path.

It presents **identically** on both the `--plugin-dir` (native plugin) and `--mcp-config` spawn paths — which confirms the problem is the MCP launcher's PATH, not plugin loading.

## Suggested fix

Pick one:

1. Resolve bun robustly in `bin/roost-irc-server` — honor a `$BUN_BIN` / `$BUN_INSTALL` override and fall back to common locations (`~/.bun/bin/bun`), mirroring the existing `ERGO_BIN` pattern.
2. Have `roost spawn` check `command -v bun` at spawn time and abort with a clear message (same shape as the zsh handling).
3. Prepend `~/.bun/bin` (or `$BUN_INSTALL/bin`) to the spawned tmux session env so the shell and the MCP inherit it.

## Workaround

`ln -s ~/.bun/bin/bun ~/.local/bin/bun` (works when `~/.local/bin` is already on the login PATH).

## Repro

Pi 5, Debian 13 trixie, aarch64. Roost v0.8.4 (also reproduces on `main` @ `8e61b1e22`). Claude Code 2.1.178. bun 1.3.14 at `~/.bun/bin`, with the bun PATH line in `~/.bashrc`. `roost spawn <project>-lead-pm --agent lead-pm …` → the agent never joins IRC; the per-project mcp-logs show `bun: not found`.

## Context

Found while bringing a roost back up; same Linux-from-source audit lineage as #517–#520. It's the bun-on-login-PATH sibling of the `~/.local/bin`-not-on-PATH note from that audit.

TeakBuilds:
Closes #599

`bin/roost-irc-server` used `exec bun` with no absolute path. When claude spawns the MCP in a non-interactive login shell (`bash -l` / `zsh -l`), only `.bash_profile`/`.zprofile` are sourced — not `.bashrc`, where the from-source bun installer adds `~/.bun/bin`. The MCP exits immediately with `bun: not found`, taking permbot with it, and the spawned agent never joins IRC.

`bin/dispatcher` has the same `exec bun` pattern and hits the same failure when the APM starts the dispatcher from within a spawned session (per `agents/associate-pm.md` line 37).

Both scripts now resolve bun via an ordered candidate list:
1. `$BUN_BIN` — explicit override; hard-errors if set but not executable
2. `$BUN_INSTALL/bin/bun` — bun installer convention
3. `$HOME/.bun/bin/bun` — default from-source install location (the actual failure path)
4. `command -v bun` — PATH fallback (homebrew, system packages)

`test/bun-resolution.test.ts` (picked up by `bun test`, ergo-gated) covers:
- **primary scenario** (the real failure): HOME=tmpdir with `.bun/bin/bun` symlinked to real bun, bun stripped from PATH, BUN_BIN unset — MCP connects to ergo via candidate-3
- **BUN_BIN override**: bun absent from PATH and HOME, BUN_BIN set — MCP connects
- **error cases**: bun not found anywhere; BUN_BIN set but not executable

### DIFF
--- a/bin/_bun-lib.sh
+++ b/bin/_bun-lib.sh
@@ -1,0 +1,48 @@
+# shellcheck shell=bash
+# _bun-lib.sh — source this; do not execute directly.
+#
+# find_bun  — resolve bun binary; prints path to stdout; exits 1 with a
+#             diagnostic on stderr if bun cannot be found.
+#
+# Resolution order:
+#   1. $BUN_BIN — explicit operator override; hard-errors if set but not executable
+#   2. $BUN_INSTALL/bin/bun — bun installer convention
+#   3. $HOME/.bun/bin/bun — default from-source install location
+#   4. command -v bun — PATH fallback (homebrew, system packages)
+
+find_bun() {
+  local candidate
+  local script
+  script="$(basename "$0")"
+
+  # Explicit operator override — hard-error if set but not executable.
+  if [ -n "${BUN_BIN:-}" ]; then
+    if [ -x "${BUN_BIN}" ]; then
+      printf '%s' "${BUN_BIN}"; return 0
+    fi
+    printf '%s: BUN_BIN=%s is not executable\n' "${script}" "${BUN_BIN}" >&2
+    return 1
+  fi
+
+  # bun installer convention: $BUN_INSTALL/bin/bun (default ~/.bun).
+  if [ -n "${BUN_INSTALL:-}" ]; then
+    candidate="${BUN_INSTALL}/bin/bun"
+    if [ -x "${candidate}" ]; then
+      printf '%s' "${candidate}"; return 0
+    fi
+  fi
+
+  # Default from-source install location.
+  candidate="${HOME}/.bun/bin/bun"
+  if [ -x "${candidate}" ]; then
+    printf '%s' "${candidate}"; return 0
+  fi
+
+  # PATH fallback (homebrew, system packages, custom installs).
+  if candidate="$(command -v bun 2>/dev/null)"; then
+    printf '%s' "${candidate}"; return 0
+  fi
+
+  printf '%s: bun not found. Set BUN_BIN to the bun binary path or install bun (https://bun.sh).\n' "${script}" >&2
+  return 1
+}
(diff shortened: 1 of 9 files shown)


bin/_bun-lib.sh
```diff
@@ -0,0 +1,48 @@
...
+  if [ -n "${BUN_BIN:-}" ]; then
+    if [ -x "${BUN_BIN}" ]; then
+      printf '%s' "${BUN_BIN}"; return 0
+    fi
+    printf '%s: BUN_BIN=%s is not executable\n' "${script}" "${BUN_BIN}" >&2
+    return 1
+  fi
+
+  # bun installer convention: $BUN_INSTALL/bin/bun (default ~/.bun).
+  if [ -n "${BUN_INSTALL:-}" ]; then
```
AlexSc, line 33: I would prefer to try path sourced before any other potential locations. It would be strange to get one bun in my shell but a different bun under roost

AlexSc (changes requested):

AlexSc (approved):

Issue #598, opened 2026-06-16
Permission classifier flattens newlines→spaces, causing false cd-multi-positional flags


TeakBuilds:
The permission safety-check classifier appears to collapse newlines to spaces before it classifies/displays a Bash command. Multi-line scripts (newline-separated statements — valid bash) get mis-parsed as a single `cd` with extra positional args and flagged `cd-multi-positional`.

Execution uses the *original* command (confirmed), so this is a classification/display bug, not an execution bug — but it generates spurious approval prompts on valid commands and shows the approver a malformed-looking command that doesn't match what the agent actually emitted.

**Example.** Agent's raw Bash tool input:

```
cd /path/to/worktree
echo "=== sandbox refs ==="
grep -rn "sandbox" src/
echo ""
find src -iname "*amazon*"
```

What the permission layer displayed + classified:

```
cd /path/to/worktree echo "=== sandbox refs ===" grep -rn "sandbox" src/ echo "" find src -iname "*amazon*"
```

→ flagged `cd-multi-positional`. Commands that put `;`/`&&` inline on the *same line* as `cd` survive the flatten and pass clean — which is the tell that newlines are the thing being lost.

**Impact:** false-positive flags on valid multi-line commands; wasted review round-trips; the displayed command misleads the approver into thinking the agent emitted broken bash (it didn't).

**Fix:** classify + display the original command string (preserve newlines), or treat newlines as statement separators in the classifier so a multi-line script isn't read as one `cd` invocation.

— filed by sdks-lead-pm (roost orchestrator), observed across multiple workers in the sdks project.

TeakBuilds:
[roost-lead-pm] Re-scoped by the #604 audit (docs/audit-2026-06-perm-irc.md, PR #609). The mechanism here is NOT a newline→space flatten — roost's `cd-multi-positional` regex uses `\s+`, and `\s` matches `\n`, so a multi-line `cd` script reads as cd-with-positional-args with no flatten step anywhere (the only flatten is `clip()` in the display path, cosmetic and downstream of classification).

[...]

PR #609, opened 2026-06-30
perm-irc ↔ Claude Code divergence audit (#604)

### LINKED ISSUE #604: validate the permissions setup: is --perm-irc requesting prompts auto-mode would auto-grant?
Investigation. Validate the whole permissions setup end-to-end, with a specific hypothesis to confirm or refute.

## Hypothesis
`--perm-irc` (the IRC permission proxy) may be firing permission requests over IRC for actions that auto permission-mode would have granted automatically — spurious prompts that add noise and friction without adding safety.

## Deliverable
Findings, not necessarily code: a clear map of where perm-irc's request behavior diverges from auto-mode's grant policy. Where it over-fires, and why. Concrete reproductions. Then a recommendation: which divergences are bugs to fix vs intended behavior.

## Why now
Pairs with #598 (perm classifier flattens newlines, causing false cd-multi-positional prompts) — that's one specific over-fire. This audit is the survey; #598 is a specific cleanup. Per our "run the survey before the paired specifics" rule, this should run first: it either confirms #598 with concrete data or re-scopes it, and may surface siblings.

Investigation issue — build on opus.

TeakBuilds:
Closes #604

Investigation deliverable: `docs/audit-2026-06-perm-irc.md`. Findings, no code change.

**Black box.** Maps where `--perm-irc`'s `classifyBash` diverges from what Claude Code 2.1.196's analyzer actually does, grounded by extracting the live classifier from the installed binary (not the stale 2.1.139 table the comments cite).

- **Hypothesis confirmed**, but the over-fires are roost regex bugs, not CC being aggressive — and the survey surfaced the opposite, more dangerous divergence too.
- **Reframe up front:** pass-through ≠ auto-grant. For a true bashMissKind, passing through hands the worker a terminal-only prompt it can't answer (hang). The lever is `emit('allow')` for the read-only subset, mirroring CC's own read-only fast-path.
- **2 over-fires (noise):** multi-line `cd` → `cd-multi-positional` (**this is #598**); `cd-compound` on read-only tails. Both roost-side.
- **4 under-fires (hang risk, higher sev):** roost misses 2.1.196 bashMissKinds `shell-expansion`, `net-redirect`, path-command `flag-validation`, redirect-target expansion → headless worker hangs. Hang prediction flagged for behavioral confirmation in the fix PR.
- **Correct mirrors (leave alone):** shell-operators (subshell/group), cd git/write/redirect, real `cd OLD NEW`.
- **#598 re-scoped:** the "flatten newlines→spaces" mechanism is wrong — it's the `cd-multi-positional` regex's `\s+` matching `\n`. Fix is split-on-separators, not "preserve newlines." O2/U1–U4 are its siblings; recommend one realignment, not one-off patches.

Reproducibility + the binary-extraction method are documented in the doc's last two sections.

### DIFF
--- a/docs/audit-2026-06-perm-irc.md
+++ b/docs/audit-2026-06-perm-irc.md
@@ -1,261 +1,218 @@
 # perm-irc ↔ Claude Code divergence audit — June 2026
 
-Investigation for #604. Hypothesis: `--perm-irc` fires IRC permission
-prompts for actions Claude Code's auto permission-mode would grant on its
-own — spurious prompts that add friction without safety. Pairs with #598
-(the cd-multi-positional specific) and gates the 0.8.6 perm batch.
+Investigation for #604. Hypothesis: `--perm-irc` fires IRC permission prompts
+for actions Claude Code's auto permission-mode would grant on its own. Pairs
+with #598 and gates the 0.8.6 perm batch.
 
-Verdict: **hypothesis confirmed, but the mechanism is narrower and more
-interesting than "the classifier is too aggressive."** Two real over-fires,
-both in roost's own `classifyBash` heuristic (not CC). The shell-operators
-seed repro does *not* reproduce as written. And the survey surfaced a more
-dangerous divergence in the opposite direction: **under-fires** where roost
-misses a CC bashMissKind, leaving a headless worker hung on a terminal prompt
-it can't see.
+This audit had two passes. The first read CC's classifier out of the 2.1.196
+binary and *inferred* behavior. Alex (#609) asked the literal operational
+question that inference can't answer, so the second pass ran real auto-mode
+Claude Code sessions and *observed* it. The observed pass is authoritative and
+is first below. The binary pass follows, corrected where observation overturned
+it — the gap between the two is itself a finding.
 
-## The headline finding: pass-through ≠ auto-grant
+## Empirical answer (authoritative)
 
-The intuitive fix for a spurious prompt is "make roost stop classifying it /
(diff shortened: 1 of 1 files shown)


AlexSc:
So my real question is "when using automode, does Claude still stop and request user permissions when using a classified bash expression?" and related "when using automode and perm-irc, does roost rely permissions requests that are unnecessary because auto mode already let the agent proceed".

Are those questions answered here? If not, that is a mandatory part of this investigation.

AlexSc (changes requested):
I want to be clear here.

We are emphatically NOT, I repeat NOT, I cannot state this enough, NOT, trying to build our own safety mechanisms or anything of the sort.

Our SOLE EXCLUSIVE GOAL is "if something that would block claude code on user input happens, make sure it gets routed over IRC so operators can be notified and unblock the agent".

So it DOES NOT MATTER IN THE SLIGHTEST that our bash classifier "doesn't catch rm -rf". If claude code does not request permissions for that, WE DON'T, END OF STORY.

That's it. Our objective here is _parity_. ABSOLUTELY NOTHING MORE, and ABSOLUTELY NOTHING LESS. We have absolutely NO INTEREST in if CC is "right" or not.

To me the important part here is that auto mode in Claude Code will bypass CC's additional bash layer of requesting. Followups need to focus on ensuring we don't generate spurious permission requests if CC is not producing blocking permission requests.

Adjust anything as needed given this.

AlexSc (approved):
The shape that we need is not loading the bash pretooluse hook when CC is in automode. That's a followup that should ideally ship in 0.8.6.

Issue #243, opened 2026-05-11
dispatcher: emit 'now watching PR/issue → routing to [channels]' event on watch-add


TeakBuilds:
Surfaced by alex on #242 review. When a PR or issue is added to watch, the operator currently has no surface that confirms which channels events from that item will route to. After-the-fact you only see it when events start firing.

Closely related to #227 (warn once on empty linked_issues): both give the operator a fast diagnostic that the watch is wired the way they intended.

## Scope

- New event kind `pr_added_to_watch_routed` (or similar — naming bikeshed welcome) routed to the project channel
- Fires on the first tick after a PR/issue appears in config's watched_prs/watched_issues
- Body: `now watching PR org/repo#N — routing events to #chan1, #chan2, ...`
- `pr_added_to_watch` event already exists per #242 worker note — currently in `SKIP_KINDS` (filtered before dispatch). Either un-skip it and add the channel list to the format, or emit a new kind.

## Acceptance

- After adding `{number: N}` to `watched_prs`, the operator sees a single confirmation in `#proj-leads` listing the channels that PR will route to
- Idempotent: doesn't refire on subsequent ticks if the watch entry is unchanged
- Same shape for `watched_issues` (separate event kind or same with a tag)

## Not in scope

- Doesn't change routing behavior, only adds confirmation
- #227 keeps warning even when explicit channels are set — that's correct because explicit channels may be additive (e.g. an extra notify channel), not substitutive for issue-derived routing

PR #618, opened 2026-06-30
route PR-scoped events to watch channel when no linked issues

### LINKED ISSUE #576: dispatcher routes review events only for PRs with linked issues; PR-watch alone insufficient
[roost-apm] from 0.8.3 milestone: PRs #570, #571, #575 (learning commits + version bump) were all missed because DM `watch pr <N>` is active but review events (APPROVED/COMMENTED/CHANGES_REQUESTED) are not relayed to `#<project>-leads` when the PR body has no `Closes #<issue>` / `Fixes` / `Resolves`.

expectation: PR-watch should relay PR-scoped events independently of issue-linking. maint PRs (version bumps, learnings, doc-only fixes) will never have a closing reference, so they always require manual polling today. fix would let the dispatcher route review + CI events to `#<project>-leads` (or a caller-specified channel) when `watch pr <N>` is active, regardless of whether the PR body has a linked issue.

### LINKED ISSUE #601: dispatcher: explicit watch channel should suppress empty-linked-issues warning and report routing instead
[carrot-lead-pm, dogfooding GoCarrot/Carrot]

When a PR is watched with an explicit channel — `watch pr <N> #proj-issue-<N>` — the dispatcher still emits the #227 empty-linked-issues warning:

> WARN PR org/repo#N has no linked issues — events won't be routed. Add `Closes #<issue>` ...

That's a false alarm here, and "events won't be routed" is flat wrong: the operator *did* specify where events route. We hit this today reviving 6 backlog PRs in GoCarrot/Carrot (#1857/#1858/#1859/#1976/#1978/#1979) — none link to a Linear issue (our PRs never do; issues live in Linear, PRs in GitHub), all six were watched with explicit `#carrot-issue-*` channels, and all six fired the warning even though routing was wired correctly.

#243 deliberately kept the warning firing even with explicit channels ("explicit channels may be additive, not substitutive for issue-derived routing"). That holds for the Linear-linked workflow, but it's the wrong default for the GitHub-PRs-don't-link-to-issues workflow, where the explicit channel *is* the substitutive route. There the warning is pure noise on every single watch, and it contradicts reality.

Pairs with #576 (PR-watch should route all PR-scoped events, incl review, to the caller-specified channel regardless of linked-issue). Once #576 lands, this warning isn't just noisy — it's misleading.

## Proposal
When a watch entry carries an explicit channel, swap the warning for the #243-style confirmation:

> routing PR org/repo#N events -> #chan (no linked issue)

## Acceptance
- `watch pr N #chan` on an empty-linked-issues PR -> no warning; one-time "routing #N -> #chan" confirmation instead
- `watch pr N` (no explicit channel) on an empty-linked-issues PR -> #227 warning unchanged
- Idempotent per #243 (doesn't refire on unchanged watch entries)

## Not in scope
- The underlying review-event routing fix — that's #576

TeakBuilds:
Closes #576
Closes #601

## What

Two routing bugs in `GitHubPrsPlugin` when a watched PR has no linked issues:

**#576** — review/CI/comment events routed to `#proj-issue-<prNum>` (a derived channel nobody watches for maint PRs) instead of the watch channel.

**#601** — `pr_no_linked_issues` emitted a misleading WARN ("events won't be routed") even when an explicit channel was specified at watch time.

## How

`prEventChannels`: drop the `#proj-issue-<prNum>` fallback. Returning `[]` when no GitHub linked issues and no Linear cross-links exist lets `resolveChannels` fall through to `entryChannels` (channels declared at `watch pr N #chan`) or `defaultChannel` (#proj-leads). All PR-scoped events now reach the watch channel regardless of linked-issue state.

`pr_no_linked_issues`: intercepted before the generic `formatPayload` path. With explicit entry channels → confirmation "routing PR X events → #chan (no linked issue)" to project channel only. Without → accurate note naming the actual routing destination. Both cases route to `[projectChannel]` only — kept separate from review/CI routing (comment in code).

Two load-bearing tests pinned: `resolveChannels([],[])` → `[defaultChannel]` (via the CI-event-routes-to-defaultChannel test) and `watch pr N` with no linked issues → events land in leads.

## Scope notes

**Supersedes #601's "warning unchanged"**: after #576, the no-explicit case routes to leads. The old "events won't be routed" would be false, so the no-explicit message is updated too. The "warning unchanged" criterion rested on the pre-#576 premise that the fallback was a spurious PR-derived channel.

**#243 interplay**: the additive concern from #243 doesn't reach `pr_no_linked_issues` — that event only fires when there are no linked issues at all, so there is no issue-channel routing for an explicit channel to be additive to. Suppressing the warning (and emitting a confirmation instead) is correct in this case.

**`#proj-issue-<prNum>` dropped from Linear-only routing too** (reviewer: this was always spurious — PR number ≠ issue number; Linear channel is the real destination).

**Single-mode foreign-repo drop fallback** now goes to `defaultChannel` instead of `#proj-issue-<prNum>`. Correct — the dropped link means no routable destination.

### DIFF
--- a/src/orchestrator/plugins/github/__tests__/plugin.test.ts
+++ b/src/orchestrator/plugins/github/__tests__/plugin.test.ts
@@ -598,6 +598,30 @@ describe('GitHubPrsPlugin.runTick — Linear attachment cross-link', () => {
     } finally { spy.mockRestore() }
   })
 
+  it('pr_no_linked_issues note names the Linear channel as routing destination when cross-link exists but no GitHub link', async () => {
+    const warnEv: OrchestratorEvent = {
+      kind: 'pr_no_linked_issues', repo: 'org/repo', pr: 25, url: 'u', title: 't',
+    } as OrchestratorEvent
+    const spy = spyOn(GhScraper.prototype, 'scrapePr').mockResolvedValue({
+      snap: fakePrSnap(), events: [warnEv],
+    })
+    try {
+      const cfg: OrchestratorConfig = {
+        project: 'proj', repo: 'org/repo',
+        plugins: {
+          'github-prs': { watched: [{ number: 25 }] },
+          'linear-issues': { watched: [{ identifier: 'C-758' }] },
+        },
+      }
+      const result = await plugin(stubFromMap({
+        'C-758': [attachment('https://github.com/org/repo/pull/25')],
+      })).runTick(cfg, { prs: {} })
+      const text = (result.taggedEvents[0]?.payload as { kind: 'oneline'; text: string }).text
+      expect(text).toContain('#proj-issue-c-758')
+      expect(text).not.toContain('#proj-leads')
+    } finally { spy.mockRestore() }
+  })
+
   it('declares Linear channels in desiredChannels for every watched Linear identifier', () => {
     const cfg: OrchestratorConfig = {
       project: 'proj', repo: 'org/repo',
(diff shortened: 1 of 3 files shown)


src/orchestrator/plugins/github/__tests__/plugin.test.ts
```diff
@@ -182,6 +182,140 @@ describe('GitHubPrsPlugin.runTick', () => {
   })
 })
 
+describe('GitHubPrsPlugin.runTick — routing when no linked issues (#576)', () => {
```
AlexSc: Remove issue reference

src/orchestrator/plugins/github/__tests__/plugin.test.ts
```diff
@@ -182,6 +182,140 @@ describe('GitHubPrsPlugin.runTick', () => {
...
+        plugins: { 'github-prs': { watched: [{ number: 25, channels: ['#proj-issue-576'] }] } },
+      }
+      const result = await new GitHubPrsPlugin('#proj-leads').runTick(cfg, { prs: {} })
+      expect(result.taggedEvents).toHaveLength(1)
+      expect(result.taggedEvents[0]?.channels).toEqual(['#proj-issue-576'])
+    } finally { spy.mockRestore() }
+  })
+})
+
+describe('GitHubPrsPlugin.runTick — pr_no_linked_issues notification (#601)', () => {
```
AlexSc: Cut issue reference

AlexSc (approved):
Nits to address before merge, rereview not needed

PR #627, opened 2026-07-01
feat: roost agents — self-updating discovery for shipped agents

### LINKED ISSUE #608: shipped agents in agents/ have no discovery path on the normal bootstrap
[roost-apm] from PR #607 / issue #605: the triage agent A1 fix added a discovery pointer (README + lead-pm "Getting started"), but the gap is structural.

Any agent shipped in agents/ is invisible on the normal operator bootstrap — claude only auto-loads .claude/agents/, and operators read README + lead-pm, not ARCHITECTURE. So they never learn a new shipped agent exists unless someone hand-patches a pointer. The general fix is a convention or small mechanism that makes every new shipped agent discoverable automatically, rather than per-agent pointer maintenance. Triage itself is now covered via the #607 fix; this is about not repeating the pattern for future shipped agents.

TeakBuilds:
Closes #608

Shipped agents in `agents/` were invisible on the normal bootstrap — claude only auto-loads `.claude/agents/`, and operators read README + lead-pm, not ARCHITECTURE. A new shipped agent only became known if someone hand-patched a pointer. This swaps per-agent pointer maintenance for one mechanism.

**`roost agents`** — two sections:
- **Shipped with roost** — the plugin `agents/` roster (name + description), discovered at runtime, so a new agent shows up on its own.
- **Spawnable here** — exactly what `roost spawn --agent` resolves from `<cwd>/.claude/agents` + `~/.claude/agents`, scope-tagged. A shipped agent missing from this set is flagged with the exact install command (`roost init --force-agents`).

**Consistency invariant:** `roost agents`' spawnable section and the `spawn --agent` not-found error are both backed by one shared helper (`_resolvable_agents`), so the roster and the error message can't drift. (The real spawn resolver in `cmd_spawn` is still a separate parallel copy of the same two-root search — identical today, but not structurally bound to the helper. Routing it through `_resolvable_agents` is a clean followup, deferred to avoid contending with in-flight work on the same function.)

**Pointers:** README and `lead-pm.md` each gain one durable pointer at `roost agents` — no edits needed when the agent set changes. That's the anti-per-agent-maintenance property the issue asked for.

**Scope note:** the command is scope-agnostic by construction — it lists whatever `agents/*.md` exists at runtime (today `lead-pm` + `associate-pm`). Triage (punted to 0.9.0) is hardcoded nowhere and will appear on its own when it lands.

**Tests:** new `test/agents_test.sh` (wired into CI) covers both sections, the reconciliation line, and the not-found↔roster consistency invariant; `spawn_test.sh` gains coverage for the enriched not-found error. `bun run lint` (incl. `shellcheck bin/*`) green; both shell suites pass (agents 7/7, spawn 34/34). No `src/` change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ roost spawn myproject-lead-pm \
 
 See [`docs/ROOST-IN-PRACTICE.md`](docs/ROOST-IN-PRACTICE.md) for the end-to-end walkthrough.
 
-Roost ships more agents than lead-pm — each is a `roost spawn --agent <name>`
+Roost ships more agents than lead-pm — each is a `roost spawn <nick> --agent <name>`
 target. `roost agents` lists the ones installed in your project; `roost agents
 --all` also shows what roost ships but you haven't installed yet, and how to
 pull them in.
--- a/README.md
+++ b/README.md
@@ -49,9 +49,9 @@ roost spawn myproject-lead-pm \
 See [`docs/ROOST-IN-PRACTICE.md`](docs/ROOST-IN-PRACTICE.md) for the end-to-end walkthrough.
 
 Roost ships more agents than lead-pm — each is a `roost spawn --agent <name>`
-target. Run `roost agents` to see the full set with descriptions, plus which
-are installed in your project (and how to pull in any that aren't). The list
-comes straight from the plugin tree, so new agents show up on their own.
+target. `roost agents` lists the ones installed in your project; `roost agents
+--all` also shows what roost ships but you haven't installed yet, and how to
+pull them in.
 
 ## Prerequisites
 
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ roost spawn myproject-lead-pm \
 
 See [`docs/ROOST-IN-PRACTICE.md`](docs/ROOST-IN-PRACTICE.md) for the end-to-end walkthrough.
 
+Roost ships more agents than lead-pm — each is a `roost spawn --agent <name>`
+target. Run `roost agents` to see the full set with descriptions, plus which
+are installed in your project (and how to pull in any that aren't). The list
+comes straight from the plugin tree, so new agents show up on their own.
+
 ## Prerequisites
 
 - macOS or Linux
--- a/skills/roost/SKILL.md
+++ b/skills/roost/SKILL.md
@@ -81,7 +81,7 @@ claude flag the wrapper doesn't otherwise know about.
 
 ## Spawnable agents (roost agents)
 
-`roost spawn --agent NAME` runs a session as that named agent. `roost agents`
+`roost spawn <nick> --agent NAME` runs a session as that named agent. `roost agents`
 lists the `NAME`s available:
 
 - **`roost agents`** — the agents installed here: what you can spawn right now.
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -48,6 +48,11 @@ roost spawn myproject-lead-pm \
 
 See [`docs/ROOST-IN-PRACTICE.md`](docs/ROOST-IN-PRACTICE.md) for the end-to-end walkthrough.
 
+Roost ships more agents than lead-pm — each is a `roost spawn --agent <name>`
+target. Run `roost agents` to see the full set with descriptions, plus which
+are installed in your project (and how to pull in any that aren't). The list
```
AlexSc: Cut sentence starting with "The list", users don't care

AlexSc (changes requested):
Let's have `roost agents` only list available spawnable agents by default. Put "could be installed, but not currently installed" behind some kind of flag. That way when lead-pm runs `roost agents` it gets a clean list of available "agents to hire" without any confusion or corner cases.

README.md
TeakBuilds: [roost-worker-608] Cut. The pointer now reads as the default/`--all` split (and lost the implementation-detail sentence).

AlexSc (changes requested):
CLI looks good. One thing to address: We ship skills/roost/SKILL.md which teaches an agent how to use the roost CLI, we should document roost agents there as well.

If you catch anything else in the skill that is behind the times, file a followup for 0.8.7.

And there's room for a durable learning here -- whenever the CLI is touched we also need to review the skill file.

skills/roost/SKILL.md
```diff
@@ -78,6 +79,21 @@ Anything passed after `--` is forwarded verbatim to claude — use this
...
+`roost agents` shows the `NAME`s you can hire:
+
+- **`roost agents`** — the agents installed here (resolved from
+  `.claude/agents/`), scope-tagged. The "who can I spawn right now" list.
+- **`roost agents --all`** — also the full roster roost ships, flagging any
+  that ship but aren't installed here plus how to install them
+  (`roost init --force-agents`). Reach for this to spot a newly shipped agent.
+
+Shipped agents land in `.claude/agents/` via `roost init`, and `roost agents`
+reads that directory — so a new one shows up without editing this skill.
```
AlexSc: "without editing this skill" is not helpful for _readers_ of this skill.

Review the docs you're writing, remember that they need to be written for an audience of _USERS_ of roost, not an audience of "people working directly on roost"

AlexSc (changes requested):

skills/roost/SKILL.md
TeakBuilds: [roost-worker-608] Fixed in ea210e4. Cut that whole sentence and re-scanned the section through a user lens — also dropped the frontmatter/model-lock detail, the `.claude/agents/` path, and "scope-tagged". The gloss now just says which command lists what you can spawn and which also shows what ships but isnt installed.

AlexSc (approved):
Just to avoid confusion, I think it's worth extending our `roost spawn --agent <name>` examples to include `<nick>`, which is still required. But I don't need to re-review after that tweak.

Make the tweaks and then you're good to merge

Issue #603, opened 2026-06-30
correct baked-in assumptions that sonnet lacks auto permission-mode


TeakBuilds:
Sonnet now supports auto permission-mode. Several places in roost bake in the old assumption that only opus gets auto mode and sonnet must use `acceptEdits`/`default`. Audit and correct them.

## Known assumption sites (audit, don't assume exhaustive)
- agent frontmatter convention: `permissionMode: auto` for opus, `acceptEdits`/`default` for non-opus (CLAUDE.md "Agent frontmatter")
- `bin/roost spawn` agent-class guidance / `--help` heuristic
- any worker/reviewer/apm spawn defaults that pick permission-mode off the model tier
- docs/LEARNINGS.md or skill text that states the opus-only-auto rule

## Deliverable
Find every place that ties auto-mode to opus specifically, decide the correct behavior now that sonnet supports it, and update consistently. This lives in 3+ surfaces, so it's a canonical-wording situation (§#422): pick the corrected sentence once, copy it verbatim. Derive the grep pattern from the known sites and scan the whole tree before editing (§#555) — the list above is examples, not an exhaustive set.

Likely sonnet, but the "find every assumption site" framing means a careful grep-the-codebase pass first.

TeakBuilds:
token cost for #603 (estimate — pricing table per-release, see src/pricing.ts):

roost-worker-603: pricing unavailable for claude-sonnet-5 — token counts only
  sonnet-5: 17k in / 62k out / 17M cache_r / 0 cache_w_5m / 202k cache_w_1h · 20m23s api / 48m04s wall

roost-reviewer-635: pricing unavailable for claude-opus-4-8 — token counts only
  opus-4-8: 8.4k in / 13k out / 654k cache_r / 69k cache_w_5m / 0 cache_w_1h · 0s api / 3m59s wall

roost-lead-pm: pricing unavailable for claude-opus-4-8 — token counts only
  opus-4-8: 32k in / 68k out / 22M cache_r / 0 cache_w_5m / 169k cache_w_1h · 23m03s api / 48m54s wall
  compaction: 1x (pre 269k → post 8.2k, 1m33s)

roost-apm: $2.59 · 8m56s api / 48m23s wall
  sonnet-4-6: 148 in / 10k out / 7.6M cache_r / 0 cache_w_5m / 27k cache_w_1h

TeakBuilds:
## Postmortem

[...]

PR #628, opened 2026-07-01
Skip bash PreToolUse hook when permission mode can't block on bash

### LINKED ISSUE #620: skip bash PreToolUse hook (classifyBash) when agent runs in auto permission-mode
[roost-apm] from PR #609 / issue #604: dont wire the bash PreToolUse hook (classifyBash / irc-pretooluse-prompt) when CC is running in auto permission mode.

In auto mode CC grants all bash without prompting, so classifyBash only ever over-fires -- producing spurious relay prompts that are never actually blocked by CC. Not loading the hook in auto mode kills the over-fire at the source and is the parity-correct fix: relay what CC blocks, nothing more.

Fix shape decided by Alex in the #609 review. #598 (cd-multi-positional regex) is a separate concern for non-auto mode and tracked independently.

TeakBuilds:
Closes #620

`--perm-irc` always wired the classifyBash/PreToolUse-Bash relay regardless of permission mode. `auto` and `bypassPermissions` never produce a blocking bash prompt in Claude Code (confirmed empirically for `auto` in the #609 audit, `docs/audit-2026-06-perm-irc.md`), so wiring the hook there only relayed noise the operator never needed to act on.

Gates hook wiring on the actual resolved permission mode: explicit `--permission-mode`, the `--model`-derived default, or — for `--agent` — a local, read-only peek at the agent's frontmatter `permissionMode:` (never passed to claude; the `--agent` path still omits `--permission-mode` entirely). Modes we can't verify skip bash prompts (no frontmatter declaration, `dontAsk`) stay conservative and keep the hook wired.

Scope: bash-specific, per the issue. `--ask-irc`/AskUserQuestion routing is untouched — no over-fire evidence there. `#598` (cd-multi-positional regex) is a separate, already-tracked concern.

## Test plan
- [x] `script/run-and-tail bash test/spawn_test.sh` — 41/41 pass, including 9 new cases covering the skip-set (auto, bypassPermissions), the conservative defaults (no declared permissionMode, `dontAsk`), the agent-frontmatter peek, explicit-flag-wins-over-frontmatter, and the AskUserQuestion scope check
- [x] `bun run lint`, `bun run typecheck` — clean
- [x] `bun test` — 992/992 pass


bin/roost
```diff
@@ -150,7 +152,12 @@ Options:
                              MCP server. It surfaces this worker's
                              tool-permission prompts as DMs to --perm-target.
                              Useful for non-Opus models that can't use auto
-                             mode. Lifecycle = MCP lifecycle.
+                             mode. Lifecycle = MCP lifecycle. The bash
+                             PreToolUse relay is skipped when the resolved
+                             permission mode never blocks on bash — nothing
+                             to relay there, so wiring it would only add
+                             noise. See cmd_spawn's skip-set for which modes
+                             qualify.
```
AlexSc: Don't put internal implementation details in user facing help text (read: cut cmd_spawn reference)

test/spawn_test.sh
```diff
@@ -519,6 +519,192 @@ fi
...
+    && ! grep -qF '"matcher":"Bash"' "$data_dir/roost-settings.json" \
+    && ! grep -qF 'hook-exec irc-pretooluse-prompt' "$data_dir/roost-settings.json"; then
+  ok "--perm-irc + opus default (auto): bash PreToolUse hook skipped"
+else
+  fail "--perm-irc + opus default (auto): bash PreToolUse hook skipped" "data_dir=$data_dir settings=$(cat "$data_dir/roost-settings.json" 2>/dev/null)"
+fi
+[ -n "$data_dir" ] && rm -rf "$data_dir"
+teardown
+
+# -- Test 29: --perm-irc + --model sonnet (acceptEdits) → bash hook still wired --
```
AlexSc: We want to switch sonnet to automode by default, so this test is covering a behavior that we intend to change

AlexSc (changes requested):
Biggest thing is that we want to switch sonnet to automode, so this is currently a hair confused

bin/roost
TeakBuilds: [roost-worker-620] Cut — help text now states the principle only, no cmd_spawn reference.

test/spawn_test.sh
TeakBuilds: [roost-worker-620] Reworked to key off explicit --permission-mode acceptEdits instead of --model sonnet's current default — this PR only asserts the skip/wire logic, not which mode sonnet defaults to. That mapping change is #603's job; per lead-pm's design intent, once #603 flips sonnet to auto the bash relay auto-skips for it with zero change here.

AlexSc (approved):

PR #634, opened 2026-07-01
add learning from #608 (doc audience)

TeakBuilds:
path-scoped to skills/**,README.md — loads when editing user-facing shipped docs.


AlexSc (changes requested):
Incorrect, README is not installed into user projects. But, it is probably wise to keep README fairly user/intro focused anyways.

AlexSc (approved):

PR #635, opened 2026-07-01
correct opus-only-auto assumption now that sonnet supports auto

### LINKED ISSUE #603: correct baked-in assumptions that sonnet lacks auto permission-mode

TeakBuilds:
Closes #603

Sonnet now supports `--permission-mode auto`, but several places baked in "only opus gets auto, everything else gets acceptEdits." This corrects the assumption to "opus and sonnet get auto; haiku (and anything unrecognized) stays acceptEdits" everywhere it lived:

- `bin/roost`: the actual model-derived default logic (`*opus*` → `*opus*|*sonnet*`), plus the `--help` prose (spawn overview, `--perm-irc`, `--permission-mode`)
- `CLAUDE.md`: the Agent frontmatter convention sentence
- `README.md` / `skills/roost/SKILL.md`: defaults section, wrapper-handles paragraph, and the `--perm-irc` primary-use-case example (now demonstrates holding a model in `acceptEdits` explicitly for oversight, since sonnet no longer needs it by default)
- `test/spawn_test.sh`: sonnet now asserts `auto`; added a haiku case to keep `acceptEdits` coverage
- `agents/associate-pm.md`: flipped `permissionMode: acceptEdits` → `auto` (the one shipped agent on the old default) — ruled in-scope by Alex; lead-pm confirmed the APM's ack-before-destructive-action behavior is a prompt-level contract independent of this value, so the merge/push approval gate is unaffected

## Test plan
- [x] `bun run lint` (eslint + shellcheck) — exit 0
- [x] `bun run typecheck` — exit 0
- [x] `bash test/spawn_test.sh` — 45 passed, 0 failed

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,7 +16,7 @@ The commands, skills, and agents this plugin ships (`prompts/`, `agents/`, `skil
 
 ## Agent frontmatter
 
-Every agent in `agents/` should declare `permissionMode:` in its YAML frontmatter — `auto` for opus agents, `acceptEdits` (or `default`) for non-opus. Claude Code reads `permissionMode:` natively on project- and user-scope agents, so the wrapper deliberately does not *pass* it — the `--agent` path of `bin/roost spawn` omits `--permission-mode` entirely and relies on the frontmatter. If a new shipped agent needs a different mode, declare it in the file — don't add a special case to the wrapper.
+Every agent in `agents/` should declare `permissionMode:` in its YAML frontmatter — `auto` for opus and sonnet agents, `acceptEdits` (or `default`) for everything else (haiku, or any unrecognized model). Claude Code reads `permissionMode:` natively on project- and user-scope agents, so the wrapper deliberately does not *pass* it — the `--agent` path of `bin/roost spawn` omits `--permission-mode` entirely and relies on the frontmatter. If a new shipped agent needs a different mode, declare it in the file — don't add a special case to the wrapper.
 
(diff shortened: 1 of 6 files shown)


AlexSc (approved):
Also we need a followup issue for 0.8.7 -- add support for Fable. Fable is even higher powered than Opus, so it should use auto permissions by default, and we need to update the pricing table as well for it

PR #642, opened 2026-07-01
Add seenBy to inbound channel messages

### LINKED ISSUE #626: Add seen by to incoming messages
Currently when an agent sends a message we include a list of nicks who saw the message, but we don't provide that information on incoming ,estates.

I've noticed agents can get confused in a group setting and believe they need to restate messages sent to group chat, apparently under the belief that the source message wasn't seen by the group.

Attaching "seen by" to incoming messages may help resolve that by directly telling agents who has seen the message sent to the channel.

TeakBuilds:
Closes #626

Incoming channel messages now carry a `seenBy` attribute (current channel member list), mirroring the seen-by hint outgoing `channel_message` replies already give (`handleSay`). Agents seeing another agent's message in a group channel can tell who else already has it, instead of assuming they need to restate it.

`buildMessageMeta` is the single builder shared by the live push path, the join-history auto-replay path, and `channel_history`'s synthesized `<channel>` tags — so all three surface `seenBy` uniformly from one change. DMs are 1:1 and skip it, same as the outgoing convention. Historical/replayed messages get a *current*-membership snapshot (not membership at original send time) — same approximation the outgoing hint already makes.

Also updated the MCP `instructions` contract string to document the new attribute.

## Test plan
- [x] `bun run lint` / `bun run typecheck` clean
- [x] New tests: live channel message carries seenBy with both nicks; DM has no seenBy; join-history replay path carries seenBy (run 3x standalone to rule out a NAMES-vs-batch-ordering race); channel_history tool's synthesized tag carries seenBy
- [x] Full suite: 995 pass, 0 fail

### DIFF
--- a/src/irc-server.ts
+++ b/src/irc-server.ts
@@ -209,7 +209,7 @@ export function createMcpServer(client: RoostIrcClient, config: ClientConfig, op
       // for to attribute a session JSONL to its roost nick. Centralized in
       // src/mcp-banner.ts so the producer (here) and the consumer can't
       // drift. Passive variant above uses a different wording on purpose.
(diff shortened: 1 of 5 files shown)


AlexSc (approved):
Nit to address before merge: Cut issue reference from tests.

Lead, can we make sure reviewer knows to look for and flag issue references in comments/code? We want to ensure the codebase is standalone and timeless. Exceptions for publicly available URLs, provided as full URLs that someone can just copy/paste into the browser.

PR #644, opened 2026-07-01
docs: fix SKILL.md dev-channels server form to match README

### LINKED ISSUE #632: SKILL.md documents dev-checkout server:roost-irc; brew install uses server:plugin:roost:roost-irc
[roost-apm] from PR #627: SKILL.md ~line 71 shows only the dev-checkout form (`--dangerously-load-development-channels server:roost-irc`). Brew-installed roost needs `server:plugin:roost:roost-irc` instead. README documents both forms; SKILL.md should match. Fix the skill to show the brew form, or both forms like README.

TeakBuilds:
Closes #632

SKILL.md's "Command surface" section hardcoded the dev-checkout form of `--dangerously-load-development-channels` (`server:roost-irc`). A brew/plugin install needs `server:plugin:roost:roost-irc` instead. README already documented both forms correctly.

Shows both forms inline, reusing README's literals character-for-character, with a pointer to README's "Debugging a failed spawn" section for the full explanation.

[roost-worker-632]

### DIFF
--- a/skills/roost/SKILL.md
+++ b/skills/roost/SKILL.md
@@ -68,7 +68,9 @@ Defaults:
   ("Agent class guidance") for which agents need this.
 
 The wrapper handles the `ROOST_IRC_*` env vars, the
-`--dangerously-load-development-channels server:roost-irc` flag, the
+`--dangerously-load-development-channels` flag (`server:plugin:roost:roost-irc`
+for a plugin/brew install, `server:roost-irc` for a bare dev-checkout MCP —
+see "Debugging a failed spawn" in README.md for detail), the
 `--permission-mode` flag (with `--agent`, defers to the agent's native
 `permissionMode:` frontmatter; with `--model`, smart-defaulted: `auto` for
 opus and sonnet, `acceptEdits` for everything else (haiku, or any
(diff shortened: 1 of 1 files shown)


skills/roost/SKILL.md
```diff
@@ -68,7 +68,9 @@ Defaults:
   ("Agent class guidance") for which agents need this.
 
 The wrapper handles the `ROOST_IRC_*` env vars, the
-`--dangerously-load-development-channels server:roost-irc` flag, the
+`--dangerously-load-development-channels` flag (`server:plugin:roost:roost-irc`
+for a plugin/brew install, `server:roost-irc` for a bare dev-checkout MCP;
+see "Debugging a failed spawn" in README.md for detail), the
```
AlexSc: Skill readers don't have our README on hand. You cannot reference it. Skill readers probably aren't in a "bare dev-checkout MCP". All of this just adds confusion to the most common case (brew install). Cut.

Lead: Can we update the reviewer to really hone in on details like this? What does _the most common user_ actually have access to/use/see? Docs need to cover that case without confusion, and _especially_ without introducing potential confusion. Can you imagine if an agent went and read the README.md for the linux kernel because they saw this line in the skill, but were in a linux kernel worktree? Confusion and chaos.

AlexSc (changes requested):

skills/roost/SKILL.md
TeakBuilds: [roost-worker-632] Fair — cut. Skill readers don't have README.md guaranteed to exist or be roost's, and dev-checkout is a maintainer-only edge case anyway. Now just shows the plugin/brew literal (`server:plugin:roost:roost-irc`), one-line swap, no cross-ref.

AlexSc (approved):

PR #649, opened 2026-07-01
Estimate api time from assistant span when turn_duration rows are absent

### LINKED ISSUE #394: roost-token-usage: 0s api time reported for short-lived reviewer agents
[roost-apm] from #roost-leads session / PR #392 cost report: reviewer made real tool calls and posted a review, but roost-token-usage reported 0s api / 3m07s wall. likely a metric collection gap for agents that exit immediately after their last turn. investigate where api time tracking is lost for short-lived agents.

TeakBuilds:
Closes #394

## Root cause

`roost-token-usage` derives api time entirely by summing `system/turn_duration`
rows. Claude Code writes those rows only at **turn completion** — when control
returns to the interactive prompt. A reviewer runs a single turn (spawn →
review → exit) and terminates at the end of that turn *before* the row is
flushed, so its transcript carries **zero** turn_duration rows and api time
reports a misleading `0s` while wall spans minutes of real work.

Confirmed empirically across five reviewers (639/642/644/645/647): 0
turn_duration rows each, same Claude Code version / entrypoint / mode as
workers. Workers, lead-pm, and apm complete turns and idle between IRC
messages, so they accumulate turn_duration normally. No reviewer row carries
any `durationMs` / `ttftMs` / result field — the timing is **structurally
absent** from the transcript, not mis-parsed. So the fix can't parse
differently; it synthesizes a fallback.

## Fix

When a transcript file has no turn_duration rows, estimate its api time from
the span of its assistant-usage timestamps, and render the nick's api figure
with a leading `~` (e.g. `~3m44s api`). For a single-turn agent that never
idles, that span is the *same quantity* turn_duration measures — busy
wall-time across the turn, tool execution included — so it's an honest
estimate, not a fudge. The `~` makes it self-documenting.

Scoping: the fallback fires only for files with **zero** turn_duration rows.
The measured path is untouched, so workers/lead/apm and long-lived agents that
lose only their in-flight final turn are unaffected. Presence of even one
in-window turn_duration row suppresses the estimate, so a forked/resumed
transcript whose rows dedup away doesn't double-count.

Verified against real transcripts:
```
roost-reviewer-639: ... · ~3m44s api / 3m44s wall   (was 0s api / 3m44s wall)
roost-worker-620:   ... · 53m12s api / 3h03m wall    (measured, no ~)
```

## Follow-up (not in this PR)

Instrumenting real api/wall at spawn time (harness-level) would give measured
numbers for short-lived agents instead of an estimate — a separate,
cross-cutting concern. Lead is filing it separately.

### DIFF
--- a/src/token-usage.ts
+++ b/src/token-usage.ts
@@ -6,6 +6,9 @@
 //     opus-4-7:    7.4k in / 63.4k out / 17.5M cache_r / 644.6k cache_w  ($14.38, miss: 128.3k ($1.42) [tools_changed 16.2k ($0.18) · system_changed 112.1k ($1.24)])
 //     sonnet-4-6:  1.2k in /  2.4k out /  300k cache_r /   8.5k cache_w  ($0.42)
 //
+// A leading `~` on the api figure (e.g. `~3m44s api`) marks an estimate —
+// see scanFile's api-time note for when and why we fall back to one.
+//
 // Sessions are matched by the MCP banner produced by src/mcp-banner.ts —
 // every roost-spawned session writes `You are connected to IRC as nick
 // "<nick>"` into its MCP instructions payload, which lands verbatim (with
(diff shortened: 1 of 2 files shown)


AlexSc (changes requested):
This is a lot of fuzzy guesswork. I propose an alternative solution: Keep reviewer agents around until the PR actually closes, establish process to have reviewer re-review after changes, and reviewer should also observe the _human_ review and draft its own learnings to try and catch things the human would in its first review pass.

PR #650, opened 2026-07-01
fix: batch GH reads into one GraphQL query per tick; split secondary rate-limit backoff

### LINKED ISSUE #613: Dispatcher: batch watches into one GraphQL call to cut API request volume
[roost-apm] from alex's request + the secondary rate-limit incident during 0.8.6 milestone work: the dispatcher round-robin polls each watch entry separately, every tick. On a busy milestone night with several active watches, this tripped GitHub's secondary (burst/abuse) rate limit — even with 2731/5000 core budget remaining.

Primary fix: batch all watch reads into a single GraphQL query so the tick costs one API call regardless of watch count. Secondary angles: honor the Retry-After header on secondary-limit 429 responses (current breaker only handles primary rate-limit), and add conditional/ETag requests so unchanged resources return 304 and don't count toward the burst budget. These complement the batch fix and together eliminate the polling pressure that caused the storm.

### LINKED ISSUE #616: dispatcher: trip + back off on secondary rate limits, honor Retry-After
[roost-apm] from PR #611: the per-entry breaker catches primary rate-limit exhaustion but not the secondary/abuse 403 that fires with primary budget still remaining. Dispatcher should detect secondary rate-limit 403s and honor the Retry-After header instead of polling straight through them. Pairs with #613's API-volume work.

TeakBuilds:
Closes #613
Closes #616

## Why

The 0.8.6 storm was burst detection, not budget exhaustion. Each 60s tick fanned ~6 REST calls per watched PR, so N watched PRs = N×6 requests — enough to trip GH's secondary (burst) limiter even with core budget to spare.

## #613 — batch reads into one GraphQL query per tick

Each plugin now issues a single aliased GraphQL query (`e0`, `e1`, …) per tick, so a tick costs **one request regardless of watch count**.

- **Per-alias isolation:** a dead entry (deleted PR, renamed repo, forbidden) comes back as a per-entry miss — its siblings still resolve, and a per-entry 404 never trips the rate-limit breaker. (`gh api graphql` exits 1 on any alias error but still writes the full `{data, errors}` body, so we parse partial success.)
- **No state migration:** snapshots stay REST-shaped — GraphQL `databaseId` maps to the numeric ids the diff keys on — so `PrSnap`/`IssueSnap` and `diff.ts` are untouched.
- **Truncation:** comment/review connections cap at the newest 100 with a visible `WARN` (not silent) if a thread exceeds it. A paged fallback is a followup if it ever bites.

### CI state via statusCheckRollup — behavior change

CI now reads `statusCheckRollup.state` (what GH's own UI shows) instead of OR-ing the REST check-runs list. **Intentional change:** a re-run PR that could stick at `FAILURE` via the old aggregate now correctly reports `SUCCESS`, because the rollup reflects the latest run per check. Mapping: `ERROR`→FAILURE, `EXPECTED`→PENDING, no-checks→null, unknown enums→null (never a false terminal). Verified against a live mixed-state PR — not 1:1 with the old aggregate, but more correct.

## #616 — split secondary rate-limit backoff

Secondary (burst/abuse) rate-limits clear in ~a minute; primary (budget) waits for the reset window. They now get separate escalating schedules:

- secondary: **1 / 2 / 5 / 10m**
- primary: 5 / 10 / 30 / 60m

The breaker notice names the kind (`GH secondary rate-limited` vs `GH rate-limited`). Classification is a best-effort stderr parse (`secondary rate` / `abuse`); anything rate-limit-shaped but untagged is treated as primary (the conservative, longer backoff). Confirmed gh does surface the secondary hint on stderr for a 403.

**Deferred to a 0.8.8 followup:** exact `Retry-After`-via-header capture. This PR ships the secondary-vs-primary split + tuned floors; reading the precise wait from the response header is the next increment.

## Cleanup

The dead REST fetch path (`fetchPr`/`fetchPr*`/`fetchIssue*` + `aggregateCi`) is removed now that the batch read replaces it.

## Tests

- Q2 isolation proven at both levels: `mapBatchOutcomes` isolates a dead alias from a crafted envelope; the plugin tick processes the sibling, carries the dead entry forward, and asserts the breaker does **not** escalate on a per-entry 404.
- rollup mapping over SUCCESS/PENDING/EXPECTED/FAILURE/ERROR/none.
- secondary-vs-primary schedules + kind-named notice, plus a whole-batch secondary integration test.
- node→snapshot mapping (databaseId→id, state lower-casing, rollup→ci_state) to pin the no-migration invariant.

`bun run typecheck` + `bun run lint` clean; `bun test` 1009 pass.


AlexSc (approved):
Because this is such a big change I'd like to see our local dispatcher cut over to this dev code now for live testing.

PR #653, opened 2026-07-01
Harden the roost agents discovery mechanism

### LINKED ISSUE #629: Harden the roost agents discovery mechanism (#608 followups)
[roost-apm] from PR #627: two low-severity robustness/DRY cleanups surfaced in the #608 review.

1. Unify the resolver: cmd_spawn's real `--agent` resolver (bin/roost ~752, `find -name "${agent}.md"`) is a third parallel copy of the same search rules that `_resolvable_agents` implements. Route the real lookup through `_resolvable_agents` (extend it to emit the path as a 3rd column) so roster, not-found error, AND the actual resolver are one search impl — fully closing the roster==resolver invariant. Deferred from #627 to avoid contending with #628's edits to the same region.

2. Folded-YAML descriptions: `_agent_description` (bin/roost ~1084) only reads single-line `description:`. A future agent using `description: >` folded/block YAML would render a mangled roster line. Make the parser tolerate (or skip cleanly) block scalars — the whole value prop is 'new agents show up on their own.'

Both are non-blocking.

### LINKED ISSUE #630: Signal when --perm-irc skips the bash relay in the spawn echo
[roost-apm] from PR #628: when the bash PreToolUse relay is skipped (resolved permission mode in the skip-set), the spawn output gives no hint — e.g. the --agent frontmatter-auto path prints `permission-mode: (claude default)` with no relay-skip signal. An operator who passes --perm-irc and expects bash prompts gets none and no explanation. Add a one-line diagnostic on the skip path (e.g. 'bash permission relay: skipped — <mode> never blocks on bash'). Low-severity operator-observability nice-to-have; out of scope for #620's core skip fix.

TeakBuilds:
Closes #629
Closes #630

Two low-severity #608 followups, combined since they're both small bin/roost robustness fixes.

- **Unify the --agent resolver**: `cmd_spawn`'s real `--agent` lookup was a third parallel copy of `_resolvable_agents`' search rules. Routed it through `_resolvable_agents` instead (now emits the file path as a 3rd column), so the roster, the not-found error, and the actual resolver read one search impl. Also makes duplicate-basename resolution deterministic (was `find | head -1`, now sorted).
- **Fold YAML block-scalar descriptions**: `_agent_description` only read single-line `description: text`. An agent using `description: >` or `description: |` block-scalar frontmatter printed a bare `>`/`|` in `roost agents --all`. The awk parser now folds block-scalar continuation lines into one line.
- **Signal the --perm-irc bash-relay skip**: when the resolved permission mode never blocks on bash (auto, bypassPermissions), the relay is intentionally left unwired — but an operator passing `--perm-irc` saw no bash prompts and no explanation. Added a one-line echo naming the resolved mode.

Tests: `test/spawn_test.sh` (+5: skip-diagnostic present/absent, duplicate-basename determinism) and `test/agents_test.sh` (+1: block-scalar folding, via sourcing bin/roost to call `_agent_description` directly since the real shipped-agents tree can't otherwise exercise a block-scalar fixture). Full suites green, lint/typecheck clean.

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -924,6 +924,11 @@ cmd_spawn() {
   esac
   if [ "${perm_irc}" -eq 1 ] && [ "${_skip_bash_hook}" -eq 0 ]; then
     pretooluse_entries+=("{\"matcher\":\"Bash\",\"hooks\":[{\"type\":\"command\",\"command\":\"${ROOST_BIN} hook-exec irc-pretooluse-prompt\"}]}")
+  elif [ "${perm_irc}" -eq 1 ] && [ "${_skip_bash_hook}" -eq 1 ]; then
+    # Otherwise an operator who passed --perm-irc expecting bash prompts sees
+    # none and has no way to tell "resolved mode never blocks on bash" apart
+    # from "something's broken".
+    echo "  bash permission relay: skipped — ${hook_perm_mode} never blocks on bash"
   fi
   if [ "${ask_routing}" -eq 1 ]; then
     pretooluse_entries+=("{\"matcher\":\"AskUserQuestion\",\"hooks\":[{\"type\":\"command\",\"command\":\"${ROOST_BIN} hook-exec irc-ask-question\"}]}")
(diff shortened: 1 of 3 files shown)


AlexSc (approved):
I think there's a long term followup issue in researching alternatives to bash for bin/roost. It's getting very complex.

Issue #11, opened 2026-05-02
Add project execution example


AlexSc:
Once I get it more hammered down, add an example of how to run planned projects using roost to manage parallelism.

PR #692, opened 2026-07-25
Batch multiple PR comments into a single IRC message

TeakBuilds:
Closes #689

When one dispatcher tick produces several comment/review events for the same channel set, the dispatcher previously sent each as a separate IRC message, letting a receiving agent start replying to one before reading the rest. This collapses consecutive `multiline`-payload events that share a channel set into one IRC message.

Black box: a tick that surfaces N inline comments (review comments + a review + a conversation comment) on one PR now arrives in the issue channel as a single IRC message with all blocks, instead of N messages. Singleton runs are unchanged.

Design:
- New `multiline_batch` payload variant (`src/orchestrator/plugin.ts`): a list of `{header, body, url}` blocks. Built only by the batcher; no plugin emits it directly.
- `batchConsecutiveMultiline` (`src/orchestrator/dispatch.ts`): pure pipeline stage merging consecutive `multiline` events with equal channel sets (sorted-set compare) into one `multiline_batch`. An intervening `oneline` event breaks the run, preserving ordering vs lifecycle/CI. Singleton runs stay `multiline`.
- `dispatchTaggedEvents` renders `multiline_batch` as one `say` per target, blocks joined by a blank line.

Granularity is channel-set, not per-PR: every comment landing in one channel in one tick is that audience's context, so cross-PR/cross-issue merges that share a channel are intended (headers disambiguate). This also makes the batcher a seam #688's backlog catch-up can reuse.

Plugin-agnostic — github-prs, github-issues, and linear all benefit. dry-run JSON keeps raw per-event payloads; only the daemon/`--dispatch-irc` render path batches.

Tests: `src/orchestrator/__tests__/dispatch.test.ts` covers the pure batcher (merge, singleton, sorted-set order-independence, split-by-channel, oneline-breaks-run, two-PRs-same-channel) and the dispatch render (one `say` per target with N blocks). Acceptance: N comments on one PR in one tick arrive as a single IRC message.

🤖 Generated with Claude Code


src/orchestrator/dispatch.ts
```diff
@@ -33,26 +33,92 @@ export async function connectAndWait(
...
       } catch (e) {
         failures.push(`${payload.kind} -> ${target}: ${e}`)
       }
     }
   }
   if (failures.length) throw new Error(failures.join('; '))
 }
+
+function renderPayload(payload: TaggedEventPayload): string {
+  if (payload.kind === 'oneline') return payload.text
```
AlexSc: Make this a switch with a default: exhaustive check. It took way too long for me to figure out that the final statement was safe and the new functionality.

src/orchestrator/dispatch.ts
```diff
@@ -33,26 +33,92 @@ export async function connectAndWait(
   await Promise.all(channels.map(ch => client.join(ch)))
 }
 
+// Normalized channel-set key for batch grouping. Sorted so order drift in a
+// plugin's resolveChannels call doesn't split a batch that should merge —
+// the audience is the set of channels, not their iteration order.
+function channelSetKey(channels: string[]): string {
+  return [...channels].sort().join('\0')
```
AlexSc, line 44: Weird way of getting the first channel, any particular reason for it?

AlexSc (changes requested):
I'm not entirely sure I understand the point of splitting on a singleton message. Can you provide an example of how that would come up and be useful?

src/orchestrator/dispatch.ts
TeakBuilds: [roost-worker-689] Done — `renderPayload` is now a switch with an exhaustive `never` default (4fb4be0). Each payload variant has a named case, and a future variant fails to compile until it adds one rather than silently falling through to `renderBatch`.

src/orchestrator/dispatch.ts
TeakBuilds: [roost-worker-689] Line 40 is `channelSetKey`, which returns a key over the *full sorted channel set*, not a first channel. The `.sort().join('\0')` builds an order-independent, collision-free grouping key: `.sort()` so `['#a','#b']` and `['#b','#a']` batch together; `'\0'` joins so `['#ab']` and `['#a','#b']` can't collide (NUL can't appear in an IRC channel name); the full set, not the first channel, because a batch can target several channels (an issue channel plus a Linear one). Added that rationale to the comment (c6d24d7).

Separately — the code that actually indexes a first event's channels, `run[0]!.channels`, lived in `batchConsecutiveMultiline`, not here. That first-event indexing is what the singleton cleanup removed: channels are now captured on a `batch` state object at the start of a run, so there's no array re-index at flush (c6d24d7).

AlexSc: I cut my teeth on C, just surprised that JS strings can natively handle nulls injected in them

TeakBuilds: [roost-worker-689] Right — JS strings are length-prefixed (UTF-16 code units), not NUL-terminated, so `'\0'` is just a regular character that sorts and joins like any other. That C instinct to read NUL as "end of string" is exactly why it's a safe separator here: it can't appear inside an IRC channel name, so `['#ab']` can't collide with `['#a','#b']`.

src/orchestrator/dispatch.ts
```diff
@@ -33,26 +33,100 @@ export async function connectAndWait(
...
+}
+
+// Collapse consecutive `multiline`-payload events that share a channel set
+// into one `multiline_batch` event, rendered by the dispatcher as a single
+// IRC message. Every comment event — including a singleton — becomes a
+// `multiline_batch` (a one-block batch renders byte-identically to a
+// `multiline` payload, so there is no separate single-comment path). Only
+// consecutive runs merge — an intervening `oneline` event breaks the run,
```
AlexSc, line 52: This is what I meant by "singleton" earlier,sorry. What's an example of an intervening one line? What does this functionally do for the end user?

AlexSc (changes requested):

src/orchestrator/dispatch.ts
TeakBuilds: [roost-worker-689] Ah — I read your earlier "singleton" as the `run.length === 1` special case (now removed; it was a no-op — a one-block batch renders identically to `multiline`, so nothing to revert). You meant the oneline-breaks-the-run behavior. Answering that:

Example: two PRs both lacking a linked issue route to `#project-leads` in one tick. PR #10 has a new review comment; PR #11 has a CI→FAILURE and a new review comment. `diffPr` emits each PR's onelines before its comment events, so the tick stream to `#project-leads` is: [PR #10 comment (multiline)] [PR #11 CI failed (oneline)] [PR #11 comment (multiline)]. The PR #11 CI line sits between the two comments. (Within a single PR this can't happen — onelines always precede that PR's comments — so it only arises across PRs sharing a channel.)

Functional effect: the agent gets three messages — [PR #10 comment] [PR #11 CI failed] [PR #11 comment] — instead of both comments in one. Comment batches stay pure (no CI line wedged inside a comment batch) and scrape order is preserved, but the two comments split around the intervening oneline, so an agent could start replying to PR #10's before seeing PR #11's — the very thing #689 is trying to prevent, in this edge.

That's a consequence of the simple consecutive-walk batcher, not a deliberate feature. The change that fully honors #689 here is to batch all comments per channel regardless of intervening onelines — but an oneline can't sit in-place inside one IRC message, so that means deferring/reordering onelines relative to scrape order (a two-pass group-by-channel). I kept the consecutive walk to preserve order and keep the change small. Happy to switch to group-by-channel if you'd rather close the edge; it's a moderate restructure, not a one-liner, so I'd run it past the PM.

TeakBuilds: [roost-worker-689] Follow-up after checking with the PM: the recommendation is to keep the consecutive-walk batcher. #689's goal is "one PR's comments arrive as one message," and within a single PR the onelines precede the comments, so that PR's comments already batch — the common case is covered. The cross-PR split is a cross-PR ordering concern rather than #689's scope, and group-by-channel would reorder/defer onelines relative to scrape order to close an edge-of-an-edge. If you'd like the cross-PR case tracked, I'm happy to have it filed as a follow-up; otherwise this PR stays as-is.

AlexSc: But if there's no intervening CI online then both PRs comment batches would come through in one IRC message?

TeakBuilds: [roost-worker-689] Correct — if there's no intervening oneline between them, PR #10's and PR #11's comments are consecutive multilines with the same channel set, so they batch into one `multiline_batch` and arrive as a single IRC message. The headers disambiguate (`PR org/r#10 comment by …` vs `PR org/r#11 comment by …`). The oneline is the only thing that breaks the run — same-channel comments from any number of PRs in one tick merge as long as nothing oneline-shaped sits between them.

TeakBuilds: [roost-worker-689] Stepping back from the keep-simple framing above — your question points at the real tension: the oneline-breaks-run is the *only* thing that splits same-channel comment batches, which is the piecemeal-reply risk #689 is after, in the cross-PR edge. So the call is genuinely yours, not a clear "keep it simple." Fair tradeoff:

- Consecutive-walk (current): simpler, preserves scrape order. An intervening oneline splits same-channel comment batches into separate messages.
- Group-by-channel (two-pass restructure): all comments routed to a channel in one tick always arrive as a single message, fully closing the split. Cost: an oneline can't sit in-place inside one IRC message, so onelines get deferred/reordered relative to scrape order, and it's a moderate restructure. The case it closes is rare — two no-linked-issue PRs hitting #project-leads in the same tick.

If you'd rather close the edge, I'll implement group-by-channel — the PM's ready to ack the restructure if you want it. If the rarity isn't worth the reorder, this PR stays as-is.

AlexSc: That's what's confusing me. Why are we special casing single line at all?

TeakBuilds: [roost-worker-689] Honest answer: there's no deliberate reason to give the oneline special treatment — it's an artifact of the single-pass, order-preserving walk. The batcher merges consecutive multilines; an oneline isn't comment-shaped, so it can't join a comment batch, and the walk flushes the batch and emits the oneline in-place. That in-place emission is what "breaks the run," and it's the only thing splitting same-channel comments.

Dropping the special-casing — batching all comments per channel regardless of intervening onelines — is exactly the group-by-channel option. The oneline then can't stay in-place (it would land inside a single IRC message that's already one comment batch), so it gets deferred/reordered relative to scrape order. That reorder is the whole cost of removing the special-case.

So: no good reason for the special-case beyond preserving scrape order. If you're fine with onelines being deferred/reordered, I drop it and do group-by-channel. Want me to?

AlexSc: I think what I'm actually hearing is that the underlying architectural issue is we have plugins emitting different "post this to the channel" events, and we've pushed "how to handle different events" into the core dispatcher, instead of having a single fixed "post to IRC" structure that the dispatcher handles simply.

Does that sound about right? TL;DR; the real tension is the dispatcher has to natively speak "pr comment block" instead of the PR plugin speaking "here's a message for IRC"?

TeakBuilds: [roost-worker-689] Yes, that's roughly right — with a nuance and a counter-weight.

Nuance: the dispatcher already spoke payload kinds before #689. The oneline/multiline split existed so it could render multiline as header/body/url vs oneline as text. #689 didn't introduce the dispatcher speaking event-shapes; it deepened it by adding the batch policy on top.

Counter-weight: the structured-payload seam is what made #689 a ~40-line change in one file, and it enables a cross-cutting merge the dumb-pipe shape can't. The tick loop concatenates every plugin's events into one list before dispatch, so the batcher merges across plugins today — a GitHub PR comment and a Linear issue comment routed to the same channel in one tick arrive as one IRC message. If plugins emitted pre-rendered "here's an IRC message" text, the dispatcher couldn't merge across plugin boundaries, and each comment-emitting plugin (github-prs, github-issues, linear-issues, and #688's backlog relay) would re-implement the batch policy. The seam exists precisely so cross-cutting dispatcher behaviors like batching land in one place.

That said, the elegant version of your idea is real: plugins emit uniform text + channel set, the dispatcher batches same-channel texts per tick and posts. The oneline/multiline distinction collapses — everything's just text lines in one channel-per-tick message, no special-casing, no reordering, no splits. Cleaner dispatcher, and arguably better for the agent (one message per channel per tick with everything that happened). The cost is the cross-plugin merge and the per-plugin policy duplication above.

So the fork:

(a) Ship focused #689 now — group-by-channel, drop the oneline special-case. Fully honors #689 (all comments to a channel → one message, always), and the batcher partially survives a later refactor (it becomes "batch all texts by channel" — a simplification, not a throwaway). Take the seam refactor as a follow-up with a deliberate design pass across all plugins.

(b) Re-scope #689 to be the refactor — move rendering into plugins, dispatcher becomes a thin uniform-text poster. Cleaner end-state, but a bigger PR that touches every plugin family, not just PR comments.

My lean is (a): ship the win now, give the refactor a deliberate cross-plugin pass rather than squeezing it through the PR-comment lens. But it's your design call — which way?

AlexSc (approved):
Go ahead and ship this now, but we need an issue for that refactor. I think that will make it easier to add more and more powerful plugins that also place nice with each other. Let's put it in an 0.9.1, but we can start dev now

PR #697, opened 2026-07-25
feat(spawn): ROOST_SPAWN_NO_LAUNCH test seam skips the real launch

TeakBuilds:
Closes #694

## What

Local-dev `bash test/spawn_test.sh` with ergo up took ~7min, and CI's `Shell tests — spawn` step took ~4m33s on main. Per-test measurement pinned two cost sources (not just the dismissal loop the issue named):

- **Unbounded `/dev/tcp` probe** — `require_ircd`'s probe to the unroutable `192.0.2.1` (2 tests) hangs ~75s each on macOS and ~136s each on Linux (TCP SYN retransmit, no timeout). This was the single biggest hit on both local-dev and CI, and it's not the dismissal loop.
- **Real claude launches (local-dev only)** — ~4 tests reaching preflight without `ROOST_SPAWN_KEEP_DATA_DIR` (tests 2, 4, 5, 18a) launched a real tmux+claude session and waited on the dev-channels dismissal loop, hitting its 30s timeout under load. CI never hit this: claude isn't installed there and `require_ircd` fails before the launch (no ergo on 6667 during the shell steps — `bun test`'s ergo is on a random port, torn down first).

## Fix

New `ROOST_SPAWN_NO_LAUNCH=1` seam in `bin/roost`: when set, spawn exits 0 right after staging (inner-cmd.txt, roost-settings.json, trust-prompt), before `require_tmux` / `require_ircd` and the real launch. `spawn_test.sh` exports it once at the top. `ROOST_SPAWN_KEEP_DATA_DIR` keeps its existing meaning (retain staged files for inspection); the two seams are independent and single-purpose.

A guard test (Test 48) pins the seam: exit 0, no `roost-testnick` tmux session left behind, returns in ≤10s, preflight staged. Without it, a broken seam falling through to `require_ircd` stays green on CI (no ergo → exit 2, every other test `||-true`s the code) and only surfaces as a slow dev box. The guard catches fallthrough on ergo-up (session left behind) and ergo-down (exit 2 instead of 0).

## Validation

- `bash test/spawn_test.sh` locally (ergo up): **6:55 → 2.2s**, 57/57 pass.
- CI `Shell tests — spawn` step: **~4:33 on main → 2s on this PR** (run 30161401641). The seam bypasses `require_ircd` entirely, so the two unroutable-IP hangs never happen on CI either.
- Ergo-up and ergo-down (dead-port) runs identical — the seam makes ergo presence irrelevant to the suite.
- Guard test verified both directions: passes with the seam intact, fails with the seam broken (ergo-up: session left behind; ergo-down: exit 2).
- `shellcheck bin/roost` clean; `agents_test.sh` (9) and `compact-hook_test.sh` (14) still pass.
- No test asserts on post-launch output, so suppressing the launch breaks nothing.

## Follow-ups

- `require_ircd`'s unbounded `/dev/tcp` probe (~75–136s hang on unroutable IPs) is a separate operator footgun — typo'd `ROOST_IRC_SERVER`. Filed as a non-blocking prod-path hardening follow-up (#696, bounded probe); PM routing via the APM.
- A coverage tradeoff around global `NO_LAUNCH` and the `require_ircd` probe path, plus an operator `--dry-run` flag, are raised on the review thread for alex.

🤖 Generated with Claude Code

### DIFF
--- a/bin/roost
+++ b/bin/roost
@@ -982,8 +982,8 @@ cmd_spawn() {
   # suppresses the failure-cleanup trap, so spawn_test.sh can read the files
   # written above after a preflight failure exits the spawn. Prod operators
   # never set this — the success path prints "data dir: ..." after tmux comes up.
-  # This does not stop the spawn: on the success path, tmux and claude still
-  # start for real — tests must tear those sessions down themselves.
+  # This does not stop the spawn: ROOST_SPAWN_NO_LAUNCH=1 (set by the test
+  # suite) is what skips the real tmux/claude launch after staging.
   if [ "${ROOST_SPAWN_KEEP_DATA_DIR:-0}" = "1" ]; then
     echo "  data dir (preflight): ${ROOST_DATA_DIR}"
   else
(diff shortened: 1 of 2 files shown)


AlexSc (changes requested):
Would a --dry-run option to spawn make sense, or is there no valid use for that?

Also ai thought CI was running ircd, tests have been taking a few minutes on Gh actions too

AlexSc:
What's with the tests setting an unroutable ip? What is the point of those?

AlexSc:
Follow-up issues: Explore/scope dry-run option

And I'm inclined to cut The whole loop back check. Operators should know what they're doing. We could add a --no-trust flag to bypass the "irc is trusted" injection. Thoughts?

AlexSc:
Sure, separate issue to add --trust-remote, 0.9.1

AlexSc (approved):

